PR #216, opened 2015-03-26
Bump admin gem to 2.1.0

fofr:
Include IP-anonymised universal analytics within support app https://github.com/alphagov/govuk_admin_template/pull/59


benilovj:
It's worth getting rid of the [GA classic tracking in the template](https://github.com/alphagov/support/blob/master/app/views/layouts/application.html.erb#L12)

PR #217, opened 2015-03-30
Upgrade to Rails 4.2.1

issyl0:
This has been going for a while and I decided to finish it off on Friday as part of 20% time, as it's been on my list for ages yet not backlog prioritised. :fearful:

Probably a good thing have done for the upcoming improvement theme, too...

(cc @benilovj)


lib/support/requests/anonymous/anonymous_contact.rb
```diff
@@ -33,6 +30,11 @@ def requester
             only_actionable.
             order("created_at desc")
         }
+
+        private
+        def anonymous_contact_params
```
benilovj: I believe that we can drop this method (as well as `ProblemReport#problem_report_params` and the other params methods) because this app only reads these models (there aren't any controllers that write these). Some time (soon, hopefully) we can update this app to read anonymous feedback through the `support-api`, not via the DB.

config/initializers/secret_token.rb
```diff
@@ -5,3 +5,4 @@
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
 Support::Application.config.secret_token = 'b806099d07d8312f4483df29a961950335c0a9c19d866b4d5f093ab6f463fe86c091abe9d5167c98a40a1f8b89765b33df5996d660061df2e3de7a4ca4725f75'
+Support::Application.config.secret_key_base = 'overwritten'
```
benilovj: we need a corresponding deployment change for this..

benilovj:
apart from the minor comments, LGTM :+1: good work!

config/initializers/secret_token.rb
issyl0: Oh, yeah. See https://github.gds/gds/alphagov-deployment/pull/874.

lib/support/requests/anonymous/anonymous_contact.rb
evilstreak: Strong parameters do the user input filtering in the controller rather than in the model. As @benilovj notes, there are no controllers actually doing that work in this app, so nowhere for strong parameters `permit` calls to go.

We can definitely drop these methods – they're unused and if they were used would error due to `params` being undefined.

evilstreak:
:+1: This is good

lib/support/requests/anonymous/anonymous_contact.rb
issyl0: Yep. Thanks for the extra explanation. I got confused between controllers and lib given what I'd done for the previous Rails upgrades was just to replace `attr_accessible` with strong_params. This has now amended 903b010.

issyl0:
Thanks @evilstreak!

issyl0:
In testing this, we discovered that something in `gds-sso` breaks because it does something Rails 4 specific and wants some new parameters, which we can't do because the support app uses a Redis store and not a User model seemingly like all the others. @benilovj can elaborate with  _much_ better words. Jake also has a commit to fix the issue we're seeing in Errbit, and one to tighten up the `gds-sso` tests to flag failing `update_attributes` operations as failing tests. I'll let him carry on... :airplane: :fr:

PR #219, opened 2015-04-16
Replace hand-rolled TablelessModel with ActiveModel::Model

benilovj:
Since the move to Rails 4.0, it's possible to use the `ActiveModel::Model`
for models that aren't backed by a database. This change removes the hand-rolled
implementation.

More info here:
http://blog.plataformatec.com.br/2012/03/barebone-models-to-use-with-actionpack-in-rails-4-0/


spec/controllers/create_or_change_user_requests_controller_spec.rb
```diff
@@ -28,7 +28,6 @@ def valid_change_user_request_params
           "email"=>"subject@digital.cabinet-office.gov.uk",
         },
         "action" => "change_user",
-        "tool_role" => "govt_form",
```
evilstreak: This change doesn't seem to match the rest of this commit. Is it intentional?

benilovj: Yes, it's intentional but I'll put it in a separate commit and add an appropriate comment

benilovj: @evilstreak I've now split this change out into d838255. The reason why it's necessary to remove that redundant field is that `ActiveModel::Model` throws a wobbly if trying to set a non-existent field.

evilstreak: Thanks for clarifying

evilstreak:
:+1:

PR #222, opened 2015-04-29
Be lenient in what the feedback explorer can receive in its URL field

fofr:
Allow full URLs with http and https, as well as partial URLs beginning with gov.uk or www.gov.uk, URLs users have entered with a bad protocol, or simply the plain path itself – with or without leading slash.

The following inputs all give the same results page:
```
Valid
https://www.gov.uk/vat-rates
http://www.gov.uk/vat-rates

Invalid
http//www.gov.uk/vat-rates
https//:www.gov.uk/vat-rates

Short hand
www.gov.uk/vat-rates
gov.uk/vat-rates
/vat-rates
vat-rates
```
Also
* removes inline form validation – what the browser validation defines as a URL is too strict.
* adds examples to the form

screen shot 2015-04-29 at 16 06 54

cc @rivalee @benilovj

### DIFF
--- a/spec/features/feedex_spec.rb
+++ b/spec/features/feedex_spec.rb
@@ -44,25 +44,7 @@ feature "Exploring anonymous feedback" do
       }
     ]
 
-    explore_anonymous_feedback_with(url: "https://www.gov.uk/vat-rates")
-    expect(feedex_results).to eq(feedback_reports)
-
-    explore_anonymous_feedback_with(url: "http:/www.gov.uk/vat-rates")
-    expect(feedex_results).to eq(feedback_reports)
-
-    explore_anonymous_feedback_with(url: "https//www.gov.uk/vat-rates")
-    expect(feedex_results).to eq(feedback_reports)
-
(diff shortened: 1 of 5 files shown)


spec/features/feedex_spec.rb
```diff
@@ -44,7 +42,28 @@
...
+    expect(feedex_results).to eq(feedback_reports)
+
+    explore_anonymous_feedback_with(url: "gov.uk/vat-rates")
+    expect(feedex_results).to eq(feedback_reports)
+
+    explore_anonymous_feedback_with(url: "/vat-rates")
+    expect(feedex_results).to eq(feedback_reports)
+
+    explore_anonymous_feedback_with(url: "vat-rates")
+    expect(feedex_results).to eq(feedback_reports)
```
benilovj: It'd be best to test this logic in a unit test, because here:

- it makes the integration test harder to read
- it distracts from the main use-case
- it makes the integration tests slower (because they run through the UI)

benilovj:
:+1:

PR #228, opened 2015-05-07
DO NOT MERGE: Use support-api

mikejustdoit:
Use an as-yet-to-be-implemented end-point of the `support-api` instead of connecting to a database.

Note: there is more work (contained in separate stories) to be done before we can completely remove the database, models, etc.

The commits here are fairly self-contained.

### DIFF
--- a/app/presenters/anonymous_feedback_presenter.rb
+++ b/app/presenters/anonymous_feedback_presenter.rb
@@ -1,0 +1,38 @@
+require 'long_form_contact_presenter'
+require 'problem_report_presenter'
+require 'service_feedback_presenter'
+
+class AnonymousFeedbackPresenter < SimpleDelegator
+  attr_reader :current_page, :total_pages, :limit_value
+
+  def initialize(api_response)
+    # actually delegate to the API response's `results` array
+    super(present_results(api_response["results"]))
+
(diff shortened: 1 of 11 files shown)


app/presenters/anonymous_feedback_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+  def presenter_for(anonymous_contact)
+    case anonymous_contact["type"]
+    when "long_form_contact"
+      LongFormContactPresenter
+    when "problem_report"
+      ProblemReportPresenter
+    when "service_feedback"
+      ServiceFeedbackPresenter
+    end
+  end
```
benilovj: I'm not sure I see the benefit of having one presenter per type at this point, given they all have the same implementation.

benilovj:
Looks good apart from the comment about the presenters.

benilovj:
@mikejustdoit apologies, I gave you bad steer about removing the JSON interface from FeedEx. I checked and it is in fact being used on an almost daily basis by the User Support analyst for ad-hoc reporting, so we can't remove it until the CSV export and date filtering is live.

benilovj:
Closing this PR because it's superseded by https://github.com/alphagov/support/pull/235.

PR #231, opened 2015-05-07
Fix label appearing twice on 'Request to unpublish content' form

benilovj:
Bump version of 'formtastic-bootstrap' gem; this picks up a bugfix (https://github.com/mjbellantoni/formtastic-bootstrap/pull/119)
which is currently unreleased to rubygems.

Before:

image

After:

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,11 @@ gem 'gds-sso', '11.0.0'
 gem 'cancan', '1.6.10'
 gem 'jquery-ui-rails', '5.0.1'
 gem 'plek', '1.10.0'
-gem 'formtastic-bootstrap', '3.1.0'
+
+# using github version to pick up unreleased bugfix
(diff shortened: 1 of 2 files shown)


mikejustdoit:
Would it be good to add a note to draw attention to this somewhere? Like a section in the README? I mean a reminder that ideally we'll upgrade to a future release at some point when one's available?

benilovj:
@mikejustdoit I can't think of an [obvious section in the README format](https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md) where it would go...

I hope that the comment in the Gemfile would be enough to however would come to pick it up. That's the natural place where I would look if I were about to upgrade gems...

mikejustdoit:
@benilovj yeah, fair point. I was just having a momentary lapse into pessimism.

PR #234, opened 2015-05-10
Improve formatting of anonymous feedback

fofr:
Make the action/problem feedback more legible by
* removing visible labels
* capitalising the start of each action/problem
* visibly rendering the action as a heading for the problem

## Before
screen shot 2015-05-10 at 10 34 50

## After
screen shot 2015-05-10 at 10 34 28

cc @benilovj @rivalee

### DIFF
--- a/app/views/anonymous_feedback/_feedback.html.erb
+++ b/app/views/anonymous_feedback/_feedback.html.erb
@@ -1,7 +1,14 @@
 <% if anonymous_contact.is_a?(Support::Requests::Anonymous::ProblemReport) %>
-  <strong>action</strong>: <%= anonymous_contact.what_doing %>
-  <br/>
-  <strong>problem</strong>: <%= anonymous_contact.what_wrong %>
+  <% if anonymous_contact.what_doing.present? %>
+    <div class="bold">
+      <span class="rm">Action:</span>
+      <%= anonymous_contact.what_doing.capitalize %>
+    </div>
+  <% end %>
+  <% if anonymous_contact.what_wrong.present? %>
+    <span class="rm">Problem:</span>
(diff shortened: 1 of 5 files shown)


benilovj:
I think that this is a good improvement.

The only thing is that we can't hide the duplicate dates until we have CSV export, because there are users who "export" FeedEx data by copying and pasting the table. Could we:

- take the removal of the date duplication into another PR
- merge the other changes her
- merge the other PR when CSV export is done?

fofr:
@benilovj I've rebased and removed date hiding, for now. A commit for the shorter date remains.

app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -1,7 +1,14 @@
 <% if anonymous_contact.is_a?(Support::Requests::Anonymous::ProblemReport) %>
-  <strong>action</strong>: <%= anonymous_contact.what_doing %>
-  <br/>
-  <strong>problem</strong>: <%= anonymous_contact.what_wrong %>
+  <% if anonymous_contact.what_doing %>
```
benilovj: You might want to do `if anonymous_contact.what_doing.present?` here to guard against `what_doing` being an empty string.

app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -1,7 +1,14 @@
...
+    <div class="bold">
+      <span class="rm">Action:</span>
+      <%= anonymous_contact.what_doing.capitalize %>
+    </div>
+  <% end %>
+  <% if anonymous_contact.what_wrong %>
```
benilovj: as above

fofr: Fixed and rebased.

PR #235, opened 2015-05-14
Change FeedEx to use the Support API to fetch feedback

benilovj:
**do not merge until**
- [x] https://github.com/alphagov/support-api/pull/48 is merged
- [x] https://github.com/alphagov/gds-api-adapters/pull/304 is merged
- [x] `gds-api-adapters` is bumped and the new version is pushed to rubygems
- [x] 5f1e309 is fixed up to depend on the newly released `gds-api-adapters` version

Before this change, FeedEx connects to the Support Contacts (MySQL) DB to fetch anonymous feedback. This is undesirable architecturally, since Support is a user-facing admin application and should be talking to an API instead.

This PR breaks the dependency on the MySQL DB; [the Support API is used as the source of the data instead](https://github.com/alphagov/support-api/pull/48).

This PR is best reviewed commit-by-commit; it roughly breaks down into the following parts:

- adjust FeedEx to read from `support-api` instead of from MySQL (2b29a42, 5f1e309, 789c5ae)
- clean up of view stuff that's no longer used (d068c3d)
- removing any last dependencies on the anonymous feedback models (39e2d41, f7280c8)
- removing DB and ActiveRecord dependencies (899d32c, a542fdd)

This PR is reuses and supersedes https://github.com/alphagov/support/pull/228.

/cc @mikejustdoit @binaryberry @boffbowsh


app/controllers/anonymous_feedback_controller.rb
```diff
@@ -23,4 +31,20 @@ def index
...
     # this is anonymous feedback, so requester doesn't need to be set
   end
+
+private
+  def index_params
+    params.permit(:path, :page)
+  end
+
+  def fetch_anonymous_feedback_from_support_api
+    api_params = { path_prefix: index_params[:path] }.tap {
```
boffbowsh: The `tap` part of this requires a bit too much reading to understand, imo:
```ruby
api_params = { path_prefix: index_params[:path] }
api_params[:page] = index_params[:page] if index_params[:page]
support_api.anonymous_feedback(api_params)
```
looks cleaner imo.

boffbowsh:
One tiny comment, otherwise it looks good :+1:

app/controllers/anonymous_feedback_controller.rb
benilovj: updated as suggested.

benilovj:
I've made changes based on @boffbowsh's comment and updated the `gds-api-adapters` as needed. This code works fine with the `support-api` locally, so this PR should be ready to merge.

fofr:
:fireworks: :tada:

evilstreak:
As a follow up to this can you update the README to include the new dependency?

benilovj:
@evilstreak good point – the README needs to be brought up to standard anyway.

PR #236, opened 2015-05-18
Add date picker

binaryberry:
Users can now pick dates from calendar. Added jQuery UI date picker, jasmine for Javascript testing. Parsing date parameters through to support-api.

cc @boffbowsh @fofr @benilovj 

## Selecting a date range
screen shot 2015-05-20 at 09 57 34
screen shot 2015-05-20 at 09 57 47

## When no feedback in date range
screen shot 2015-05-21 at 10 40 28

## Open ended date ranges and lenient date formats
screen shot 2015-05-20 at 09 58 04
screen shot 2015-05-20 at 09 58 32
screen shot 2015-05-20 at 09 58 16

## Warnings when filter fails
screen shot 2015-05-21 at 10 42 31

### DIFF
--- a/app/helpers/date_filter_helper.rb
+++ b/app/helpers/date_filter_helper.rb
@@ -1,7 +1,23 @@
 module DateFilterHelper
 
 	def date_filtered?
-		params[:from].present? || params[:to].present?
+		@from_date.present? or @to_date.present?
+	end
+
+	def total_responses_header(total_count, from, to)
+		response_total = pluralize(number_with_delimiter(total_count), 'response')
+
+		if from.present? and to.present?
+			"#{response_total} between #{from} and #{to}"
+		elsif from.present?
+			"#{response_total} since #{from}"
+		elsif to.present?
+			"#{response_total} before #{to}"
+		elsif total_count && total_count > 1
+			"All #{response_total}"
+		else
+			response_total
+		end
 	end
 
 end
--- a/app/helpers/date_filter_helper.rb
+++ b/app/helpers/date_filter_helper.rb
@@ -4,6 +4,14 @@ module DateFilterHelper
 		@from_date.present? or @to_date.present?
 	end
 
+	def attempted_to_filter?
+		params[:from].present? or params[:to].present?
+	end
+
+	def invalid_filter?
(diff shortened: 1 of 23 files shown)


fofr:
This is working nicely. :+1:

I have some tweaks to suggest around the way the filter/URL is working with partial date ranges and custom date formats – I think user input that doesn't match the specified format, or which provides only partial information should be coerced into the correct format.

The feedback order (reverse-chronological) when filtered feels counterintuitive.

We're also missing an easy way to to remove the filter which I forgot to include.

These are mostly kinks that a non-functional prototype wouldn't catch.

benilovj:
A small comment: the last two commits were made by the vagrant user; it'd be best to amend those with proper author data, and set up git within the VM with your credentials...

boffbowsh:
Thanks for adding screenshots, @binaryberry

app/helpers/date_filter_helper.rb
```diff
@@ -0,0 +1,31 @@
+module DateFilterHelper
```
benilovj: indentation doesn't look quite right on this module

benilovj:
Apart from minor style remark, :+1:

app/helpers/date_filter_helper.rb
binaryberry: True! Sorted now.

benilovj:
:+1:

app/helpers/date_filter_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+  def total_responses_header(total_count, from, to)
+    response_total = pluralize(number_with_delimiter(total_count), 'response')
+
+    if from.present? and to.present?
+      "#{response_total} between #{from} and #{to}"
+    elsif from.present?
+      "#{response_total} since #{from}"
+    elsif to.present?
+      "#{response_total} before #{to}"
+    elsif total_count && total_count > 1
```
tadast: Is there a reason why there are both `and` and `&&` here? :)

binaryberry: Not really @tadast ! This PR has been merged now but I'll turn the "and" to "&&" in another branch we are working on. Consistency IS good :-)

PR #237, opened 2015-05-19
Stub requests in specs

mikejustdoit:
A small clean up to some of the specs' use of GdsApi request stubs.

A couple of these issues are caused by WebMock's need to have stubs explicitly cleared after use - i.e. later specs were relying on API request stubs from earlier specs. (We could try to explicitly clear stubs after use in a few different ways, if it was agreed to be worthwhile. An alternative is running the specs in a random order, as I did to find these problems this time.)

I haven't managed to get to the bottom of one of these issues. It seemed as though an API request stub from an earlier spec was being triggered from a later spec that had already performed its own request stub with the same URI. For now I've just changed their URIs to be unique. That is clearly not a long-term solution.


benilovj:
Is it worth doing `WebMock.reset!` in the teardown of some specs?

PR #238, opened 2015-05-21
Remove initialiser for PP

benilovj:
The app no longer talks to the PP (all uploads happen in support-api).


fofr:
I'm guessing PP is Performance Platform?

fofr:
Presumably there'll be a corresponding PR in `alphagov-deployment`? Otherwise :+1:

benilovj:
@fofr  see https://github.gds/gds/alphagov-deployment/pull/915

PR #239, opened 2015-05-22
Prevent invalid query strings in referrers from erroring

fofr:
For certain referrer URLs  eg `Taxed%20to%2`, `Rack::Utils.parse_query(uri.query)` throws an ArgumentError: `invalid %-encoding`.

Because this is mostly an enhancement to the table, rather than attempt to decode the query correctly, return a nil search term.

Interestingly, before Ruby 2.1 this URL would have thrown `URI::InvalidURIError` and the error would have been caught.

A discussion on the nuances of this issue can be found here:
https://github.com/rack/rack/issues/337

cc @benilovj @boffbowsh


boffbowsh:
LGTM :+1:

PR #241, opened 2015-05-22
Add support for filtering orgs by acronym

boffbowsh:
Break link to Whitehall organisations API, preferring support api organisations list
Add the status of the organisation to the list if it’s not live
Default the selected organisation to the currently logged-in user's organisation

### DIFF
--- a/app/controllers/anonymous_feedback/explore_controller.rb
+++ b/app/controllers/anonymous_feedback/explore_controller.rb
@@ -1,15 +1,18 @@
 require 'support/requests/anonymous/explore'
-require 'gds_api/organisations'
+require 'gds_api/support_api'
 
 class AnonymousFeedback::ExploreController < AuthorisationController
   authorize_resource class: Support::Requests::Anonymous::Explore
 
   def new
     @explore_by_url = Support::Requests::Anonymous::ExploreByUrl.new
-    @explore_by_organisation = Support::Requests::Anonymous::ExploreByOrganisation.new
-    @organisations_list = organisations_api.organisations.map { |org|
-      [org.title, org.details.slug]
-    }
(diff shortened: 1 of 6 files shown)


app/controllers/anonymous_feedback/explore_controller.rb
```diff
@@ -1,15 +1,18 @@
...
   authorize_resource class: Support::Requests::Anonymous::Explore
 
   def new
     @explore_by_url = Support::Requests::Anonymous::ExploreByUrl.new
-    @explore_by_organisation = Support::Requests::Anonymous::ExploreByOrganisation.new
-    @organisations_list = organisations_api.organisations.map { |org|
-      [org.title, org.details.slug]
-    }
+    @explore_by_organisation = Support::Requests::Anonymous::ExploreByOrganisation.new(organisation: current_user.organisation_slug)
+    @organisations_list = support_api.organisations_list.map do |org|
```
benilovj: this has quite a lot going on - maybe pull out both org list construction, and org title construction into separate methods?

benilovj:
:+1:

edds:
I know this has already been merged but what was the reasoning behind "Break link to Whitehall organisations API, preferring support api organisations list"? The commit message doesn't make it clear as to why you would use one over the other.

benilovj:
@edds as discussed in person (and for the benefit of anyone doing GitHub archeology), we've decided to fetch organisations from the Organisations API offline in `support-api`, rather in `support` (in this context, `offline` means that there's a nightly cron that imports orgs into the `support-api` DB). This is mainly to do with the fact that currently the Organisations API is a bit slow when uncached, although as @edds pointed out, on prod the responses are cached for 15 minutes.

PR #245, opened 2015-05-27
[DO NOT MERGE] Added Universal Analytics tracking for date picker

binaryberry:
[DO NOT MERGE] - in the process of implementing oral feedback from Jake Benilov!

eventAction is tracking: whether date ranges provided 'to' date, 'from' date, neither or both.
eventLabel is tracking:
1) length of date range chosen by users, which will later help decide on preset time filters to add to feedex.
2) time between search and 'to' date provided. If the user does a search of all anonymous contacts  until today, it will return 0.

Added script that will run in development to display the values sent to Google Analytics. In production, this will be performed by the [gov_uk_admin_template](https://github.com/alphagov/govuk_admin_template) gem.

screen shot 2015-05-27 at 14 41 42

cc @fofr @benilovj @boffbowsh

### DIFF
--- a/app/helpers/date_filter_helper.rb
+++ b/app/helpers/date_filter_helper.rb
@@ -12,7 +12,10 @@ module DateFilterHelper
     attempted_to_filter? && !date_filtered?
   end
 
-  def total_responses_header(total_count, from, to)
+  def total_responses_header(total_count, from_date, to_date)
+    from = from_date.to_s(:govuk_date_short) if from_date
+    to = to_date.to_s(:govuk_date_short) if to_date
+
     response_total = pluralize(number_with_delimiter(total_count), 'response')
 
     if from.present? and to.present?
@@ -26,5 +29,26 @@ module DateFilterHelper
     else
       response_total
     end
+  end
+
+  def date_range_track_action(from_date=nil, to_date=nil)
+    if from_date && to_date
+      "filtering between two dates"
+    elsif from_date
+      "filtering everything since date"
+    elsif to_date
+      "filtering everying before date"
+    else
+      "no filtering"
+    end
+  end
+
+  def date_range_track_label(from_date, to_date)
+    from_date ||= Date.new(1970,1,1)
+    to_date ||= Date.current
+    date_range = distance_of_time_in_words(from_date, to_date)
+    date_ago = time_ago_in_words(to_date)
+    date_ago = 0 if to_date.at_middle_of_day() == Date.current.at_middle_of_day()
+    return "selected date range: #{date_range}, time between search and 'to' date provided: #{date_ago}"
   end
 end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -39,6 +39,13 @@
   </div>
 
   <%= yield %>
+    <% if Rails.env.development? %>
+      <script>
+        window.ga = function() {
+          console.log(arguments);
+        }
+      </script>
+    <% end %>
 <% end %>
 
 <% content_for(:footer_version) { CURRENT_RELEASE_SHA } %>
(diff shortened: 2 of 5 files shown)


app/helpers/date_filter_helper.rb
```diff
@@ -27,4 +30,25 @@ def total_responses_header(total_count, from, to)
...
+    end
+  end
+
+  def date_range_track_label(from_date, to_date)
+    from_date ||= Date.new(1970,1,1)
+    to_date ||= Date.current
+    date_range = distance_of_time_in_words(from_date, to_date)
+    date_ago = time_ago_in_words(to_date)
+    date_ago = 0 if to_date.at_middle_of_day() == Date.current.at_middle_of_day()
+    return "selected date range: #{date_range}, time between search and 'to' date provided: #{date_ago}"
```
fofr: We’ll want to shorten this label so that it's scannable in the analytics tool.

@boffbowsh Is this including the data that you expect?

app/views/layouts/application.html.erb
```diff
@@ -39,6 +39,13 @@
   </div>
 
   <%= yield %>
+    <% if Rails.env.development? %>
+      <script>
+        window.ga = function() {
+          console.log(arguments);
```
edds: Can this check for the existence of console.log? Otherwise this will never work in IE in development, which we should be checking when making front end changes.

fofr: I don't think this script should be included at all.

app/helpers/date_filter_helper.rb
```diff
@@ -27,4 +30,25 @@ def total_responses_header(total_count, from, to)
       response_total
     end
   end
+
+  def date_range_track_action(from_date=nil, to_date=nil)
```
benilovj: As mentioned in person, it's probably worth extracting `date_range_track_action` and `date_range_track_label` into their own class (eg something like):

```ruby
class DateRangeTrackEvent
   def initialize(from_date, to_date)
     ...
   end

  def action
    ...
  end

  def label
    ...
  end
end
```

edds:
Been thinking about this overnight. Is this needed? While putting the data in events does give you an easier way of getting the data out, you should already be able to get this data from google analytics without using events as the data is being put into the page query string. For example [here are all the different date filters people have used since the first of the month](https://docs.google.com/a/digital.cabinet-office.gov.uk/spreadsheets/d/1kCPPWHdG4kBhRApbAQ7V1-oO2tk_Bucr9P-0O6zP340/edit?usp=sharing).

app/helpers/date_filter_helper.rb
boffbowsh: I think these need to be two discrete metrics being logged, perhaps as separate actions. I really don't know enough about what's possible for data extraction in GA, I think we need to properly research what's actually possible and discuss with Syed. I think we need to look at the existing GA account and see examples of this sort of data being extracted, and check what's possible.

fofr:
@edds Something that that export doesn't give you (without some extra work, though it is possible) is the day on which they filtered. We want to use this data to determine if there are any common date presets that might be useful, eg filtered May 5 to May 6, but did so on May 7 – that would be a common filter of the last two days, for instance.

(Research with users has shown that reports they generate are often over date ranges unique to their organisation.)

The story is here: https://tree.taiga.io/project/core-improvement-theme/us/53?kanban-status=211660 (but that also lacks detail)

edds:
@fofr adding date to the export is just a case of adding `ga:date` to the dimensions. I have taken the liberty of adding that to that spreadsheet I created. That now gives you the dates they filtered by and the date they did that filtering.

fofr:
Nice :+1:

PR #248, opened 2015-06-05
Filter Problem Reports by Organisation

boffbowsh:
Presents the work in https://github.com/alphagov/support-api/pull/60 to the user.

Their are some [TODO notes](https://github.com/alphagov/support/compare/alphagov:7ce7e5e...alphagov:7cdf11e#diff-ce355e2e49a16102ff6d28143466fa8eR28) left by other members of the team. We'll address these when the team is back together next week.


app/controllers/anonymous_feedback_controller.rb
```diff
@@ -36,17 +47,54 @@ def set_requester_on(request)
...
+  def fetch_anonymous_feedback_from_support_api
+    AnonymousFeedbackApiResponse.new(
+      support_api.anonymous_feedback(api_params).to_hash
+    )
+  end
 
-    support_api.anonymous_feedback(api_params)
+  def organisation_title(slug)
+    org = support_api.organisations_list.detect { |o| o["slug"] == slug }
+    org.present? ? org["title"] : slug
```
mikejustdoit: Maybe we could get the organisation title back with the API response instead?

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -6,27 +6,38 @@ class AnonymousFeedbackController < RequestsController
...
+        @feedback = AnonymousFeedbackPresenter.new(api_response)
+        @dates = present_date_filters(api_response)
+        # TODO: we should decide how to describe filtering by both a path and
+        # an organisation, separately and together
+        # TODO: I guess we should determine this filtering information from the
+        # api_response rather than user-supplied params
+        # TODO: this shouldn't be here, it belongs in its own object, possibly
+        # combined with date filters in a more general filters presenter?
+        if index_params[:path].present?
+          @filtered_by = index_params[:path]
```
mikejustdoit: Would it be nicer if we were working with the `path` and/or `organisation` from the API response instead of user-supplied params here? (As described in that TODO just above.)

app/controllers/anonymous_feedback_controller.rb
boffbowsh: That doesn't feel like a good API design. "Please give the feedback for this organisation. Also tell me what the organisation is named". Linked too much to use-case. I'd rather add a separate API call for info about an org, but the obvious endpoint is used by the organisation summary stuff.

mikejustdoit: Hmm, yeah, good point.

If that existing endpoint isn't correct, we should change it

boffbowsh: `/anonymous-feedback/organisations/:id` is currently the summary. Maybe we could have `/organisations/:id` for the data about the organisation? And also move `/anonymous-feedback/organisations` as the list of orgs for autocomplete purposes to `/organisations`? It's all really just a hack due to the lack of an Organisations API that's quick enough :/

mikejustdoit: You're right, it does seem a bit of a shame to be reinventing it.

Since we've already got the Organisations list, then yes, if it's not too much work to have those two related resources at `/organisations/` and `/organisations/:id`, I think it's the best we can do for now.

boffbowsh: Work done here https://github.com/alphagov/support-api/pull/61

app/controllers/anonymous_feedback_controller.rb
boffbowsh: I don't like that idea - why should the support API return the parameters just to support this use case?

mikejustdoit: My bad, I assumed the API already did return them. I guess since it returns the from and to dates.

bishboria:
:+1:

PR #250, opened 2015-06-08
Ensure we can filter by date for organisation

boffbowsh:
Missed this. Need to add much better test coverage for this functionality.


bishboria:
:+1:

PR #255, opened 2015-07-22
Prevent empty URL searches from erroring

fofr:
When a URL is not present, the controller would render the `new` template but without going through the new method it wouldn’t have all the template variables it needed.

* Call the `new` method before rendering with a 422 status code

https://trello.com/c/OAJAO2RB/


issyl0:
Thanks!

PR #258, opened 2015-09-22
Move dates to comment body

jamiecobbett:
By moving the dates to the comment body, they're always displayed. At the moment, they're only displayed if the user turns on "Show all events".

This also includes a refactoring to use the TimeConstraint code for ContentAdviceRequests, which builds on the first commit. This second bit could be split off if needed.


jamiecobbett:
@benilovj quite right - see https://github.com/alphagov/gds_zendesk/pull/23

PR #263, opened 2015-10-14
Add training information to create or change user

binaryberry:
When deciding on granting user permissions, people need to know what previous training requesters have had.

Discussed design and copy with @fofr and Nick Cammel from the content team.

before
screen shot 2015-10-14 at 16 31 54

after
screen shot 2015-10-14 at 16 31 22

cc @fofr @boffbowsh 

[Trello](https://trello.com/c/pi5NbNsj/119-update-the-create-or-change-a-user-account-form-in-support-app-small)


fofr:
Why does the copy now say "you"? Shouldn't the training relate to the user the request is for?

PR #264, opened 2015-10-15
Harmonise copy

binaryberry:
Change copy to reflect the fact that the person who fills in the form
is requesting changes for someone else.

cc @fofr 

[Trello](https://trello.com/c/pi5NbNsj/119-update-the-create-or-change-a-user-account-form-in-support-app-small)


boffbowsh:
Should be "GOV.UK training"

binaryberry:
Thanks! Amended now! :rainbow:

PR #269, opened 2015-11-06
UI changes to support feedex result limiting.

boffbowsh:
Uses the `results_limited` flag from support-api to display more
accurate response counts if the user’s query has been truncated due to
too many results. Also displays useful information on the last
available page.

Before:

screen shot 2015-11-06 at 14 35 16

After:

screen shot 2015-11-06 at 16 01 34

Needs pointers from @fofr and Gavan on layout and copy.


fofr:
The way I'd imagined this working was:
* Using the copy "Over 10,000 responses"
* Showing a message about finding more responses when a user reaches the end of the list, ie on the last page.

boffbowsh:
Updated to show useful copy on the last page only, and changed "Most recent" to "Over".

PR #270, opened 2015-12-08
Upgrade GDS SSO gem to v11.2.0

benilovj:
This involved adding a missing method to the User model. The main
motivation is to stay up-to-date with GDS SSO.

/cc @benlovell


benlovell:
:sparkles:

PR #273, opened 2016-03-14
Prevent duplicate Zendesk ticket creation

binaryberry:
When receiving a 409 response (conflict) from the Zendesk API, the support app keeps trying to create a ticket until the request goes through. 

This change rescues the 409 response and takes no action, ignoring the ticket creation because Zendesk has already received it. 

This PR also desactivates the submit button in the view after it has been pressed, to prevent duplicate requests being generated in the frontend (the below would usually be displayed for less than a second).

screen shot 2016-03-14 at 17 52 34

Another change that was considered was to store a hash of the ticket requester/contents for a few minutes in an ephemeral cache (eg redis) and drop any duplicates - however we think the two changes above are enough to solve the problem (do jump in @benilovj if you disagree).

[Trello ticket](https://trello.com/c/dnYu5sNT/312-fix-zendesk-duplicate-ticket-issue-medium)


app/workers/zendesk_ticket_worker.rb
```diff
@@ -1,11 +1,16 @@
 class ZendeskTicketWorker
   include Sidekiq::Worker
-  
+
   def perform(ticket_options)
     if GDS_ZENDESK_CLIENT.users.suspended?(ticket_options["requester"]["email"])
       $statsd.increment("#{::STATSD_PREFIX}.report_a_problem.submission_from_suspended_user")
     else
-      GDS_ZENDESK_CLIENT.ticket.create!(HashWithIndifferentAccess.new(ticket_options))
+      create_ticket(ticket_options) rescue ZendeskAPI::Error::NetworkError
```
benlovell: Maybe we should be logging this, rather than swallowing it inline? I'm guessing `NetworkError` can represent a wider range of failure modes?

spec/workers/zendesk_ticket_worker_spec.rb
```diff
@@ -18,4 +18,15 @@
 
     expect(a_request(:post, %r{.*/tickets/.*})).to_not have_been_made
   end
+
+  it "discards the ticket if it receives a 409 response" do
+    zendesk_has_user(email: "a@b.com", "suspended" => false)
```
benlovell: Is there a reason for these options to be a mixture of 1.9 and hash rocket styles?

benilovj:
I don't think that the duplicate submission protection on the government support request forms will help here. Almost all of the duplicates originate through the [public support form](https://www.gov.uk/contact/govuk), which go through the `feedback` app. I'm not sure we'd see any benefit from having the protection there, given the low occurrence of government support request duplicate tickets.

On the broader question of "is discarding the 409 response enough?", my previous analysis suggested that it may not be enough, but we cannot be sure until we make this change. I'd suggest that we come back in a month after releasing this change and see how many duplicates there are now, and decide whether further duplicate detection is necessary. ^ @boffbowsh

app/workers/zendesk_ticket_worker.rb
boffbowsh: Indeed, we need to check that the error is a 409. We'd want to retry a 5xx.

spec/workers/zendesk_ticket_worker_spec.rb
```diff
@@ -18,4 +18,15 @@
 
     expect(a_request(:post, %r{.*/tickets/.*})).to_not have_been_made
   end
+
+  it "discards the ticket if it receives a 409 response" do
```
boffbowsh: Should add a spec for a 5xx too

boffbowsh:
Now that I've had :coffee: I understanding what Jake means. This change probably belongs more in the `feedback` app than in here.

spec/workers/zendesk_ticket_worker_spec.rb
binaryberry: The reason was being consistent with other tests on this page. But I've now amended them too in 74d59133.

binaryberry:
Ok, thanks Jake! I didn't realise the duplicates came from the public support form. I've now removed that commit from the PR, and put some info about it in Trello [here](https://trello.com/c/3aTwn8Lm/509-review-feedex-duplicates-bug) in case it can help for changes to the feedback app later.

spec/workers/zendesk_ticket_worker_spec.rb
binaryberry: Sure - done in 0f0e3d9.

app/workers/zendesk_ticket_worker.rb
binaryberry: Thanks both! Addressed in 0f0e3d9.

app/workers/zendesk_ticket_worker.rb
```diff
@@ -1,11 +1,20 @@
...
+
   def perform(ticket_options)
     if GDS_ZENDESK_CLIENT.users.suspended?(ticket_options["requester"]["email"])
       $statsd.increment("#{::STATSD_PREFIX}.report_a_problem.submission_from_suspended_user")
     else
-      GDS_ZENDESK_CLIENT.ticket.create!(HashWithIndifferentAccess.new(ticket_options))
+      begin
+        create_ticket(ticket_options)
+      rescue ZendeskAPI::Error::NetworkError => e
+        raise unless e.response.status == 409
```
benilovj: When the response status isn't `409`, this statement raises a `RuntimeError`, rather than re-raise the original network error. I think it'd be better to re-throw the original error, as it contains all the context on it.

benilovj: I think it may also be worth adding a comment in the code to explain what makes the 409 status code so special in this case, but that's subjective...

binaryberry: Sure! Added [here](https://github.com/alphagov/support/pull/273/files#diff-d0ae5436575438ec080433ef967a65cdR11) now.

benilovj:
This looks good. One final thing; we discard the exception completely silently, and there's no way to track that it's happened. Is it worth having a stats counter to track these (something like [this](https://github.com/alphagov/support/blob/e6898c6858b55aa578787fdefe140a5206fc26bf/app/workers/zendesk_ticket_worker.rb#L6)), in case Zendesk tickets start going missing and we don't know why?

binaryberry:
Oki doki! Adressed [here](https://github.com/alphagov/support/pull/273/files#diff-d0ae5436575438ec080433ef967a65cdR12).

benilovj:
:+1:

PR #275, opened 2016-04-06
Default feedback query to last 30 days

boffbowsh:
In order reduce expensive queries for feedex, by default only show feedback from the
last 30 days. Users can still change the date range once the initial query has loaded.

Added a new test stub helper so as not to pollute the tests too much with this behaviour.

/cc @benilovj


fofr:
What was the previous default?

boffbowsh:
Everything ever

PR #277, opened 2016-05-27
Rollup ratings

binaryberry:
This presents aggregated feedback generated from [this support-API PR](https://github.com/alphagov/support-api/pull/82).

Two other commits to stop a test from failing due to a whitespace error and to stop a warning about the raise_error matcher.

[Trello](https://trello.com/c/zhqhIKOs/377-reduce-duplication-of-quantitative-results-in-feedex-medium)


app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -15,6 +15,13 @@
     <br/>
     <strong>comment</strong>: <%= anonymous_contact.details %>
   <% end %>
+<% elsif anonymous_contact.type == "aggregated-service-feedback" %>
+  <strong><%= anonymous_contact.service_satisfaction_rating %> rating</strong>:
```
boffbowsh: "rating" should be plural imo but it might be good to check the microcopy with someone like @lutgendorff

app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -15,6 +15,13 @@
...
+  <% if anonymous_contact.details.present? %>
```
boffbowsh: If this isn't present, it'll just render an empty string. So you can just get rid of the if block guarding it.

app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -15,6 +15,13 @@
...
+    <%= anonymous_contact.details %>
+  <% end %>
+<% elsif anonymous_contact.type == "service-feedback-rating" %>
```
boffbowsh: Actually looking at it I think these if statements are the wrong way round? This should be the aggregate, and the above should be the ones with comments?

binaryberry: Urgh, yes. Thanks for spotting this.

app/views/anonymous_feedback/_feedback.html.erb
binaryberry: done!

app/views/anonymous_feedback/_feedback.html.erb
binaryberry: done, Liz is out until Monday I think but could maybe check with a content person?

app/views/anonymous_feedback/_feedback.html.erb
```diff
@@ -15,6 +15,11 @@
     <br/>
     <strong>comment</strong>: <%= anonymous_contact.details %>
   <% end %>
+<% elsif anonymous_contact.type == "service-feedback-rating" %>
```
boffbowsh: Is this different to the block above? The types haven't changed have they?

boffbowsh:
This is missing the "Remove 30 day default filter" acceptance criteria

app/views/anonymous_feedback/_feedback.html.erb
binaryberry: No, sorry that was a version control issue that also impacted how the service feedback aggregate details were called, but amended now. Tweaked the wording too following Liz's feedback, cf. updated screenshot in https://github.com/alphagov/support-api/pull/82 .

binaryberry:
@boffbowsh adressed now!

PR #279, opened 2016-07-05
Remove headers argument

mgrassotti:
The additional argument was added to solve a problem with the new
govuk_sidekiq gem.

Now the argument is correctly removed by the gem govuk_sidekiq/testing module.


boffbowsh:
👍  merge on Thursday

PR #280, opened 2016-07-13
Remove responses count for done pages

binaryberry:
Pages with a path starting with "/done/" have another type of anonymous contact, in addition of comments, called service satisfaction ratings. Due to high volume, the latter are rolled up each evening into service rating aggregates that show the count of each rating. Each service rating aggregate is added as a line in the feedex table, and service satisfaction ratings that have no comments are erased from the table. 

This means that the number of "responses" displayed at the top of the table is misleading for these /done/pages, as they are in fact the number of lines in the table, which for /done/ pages is not relevant. For the feedback of the present day, there is no aggregation, so the number of lines in the table is the number of service satisfaction ratings. For the feedback of previous days, the number of lines in the table is the number of service satisfaction ratings with comments + number of service rating aggregates (zero to five lines). So it's not a very useful number.

Some users have found it confusing and misunderstood it, so this change removes the responses count altogether for paths starting with "/done".

Before
screen shot 2016-07-13 at 16 57 27

After
screen shot 2016-07-13 at 17 02 36

Paired on this with @mgrassotti .

[Trello](https://trello.com/c/TVlBgbol/425-feedex-aggregation-not-adding-up-correctly)

### DIFF
--- a/app/helpers/feedex_helper.rb
+++ b/app/helpers/feedex_helper.rb
@@ -1,8 +1,7 @@
 module FeedexHelper
-  def total_responses_header(total_count:, from:, to:, results_limited:)
-    response_total = pluralize(number_with_delimiter(total_count), "response")
+  def total_responses_header(total_count:, from:, to:, results_limited:, path:)
 
-    response_total = "Over #{response_total}" if results_limited
+    response_total = get_response_total(total_count, results_limited, path)
 
     if from.present? && to.present?
       "#{response_total} between #{from} and #{to}"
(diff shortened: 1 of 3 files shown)


app/helpers/feedex_helper.rb
```diff
@@ -12,8 +11,22 @@ def total_responses_header(total_count:, from:, to:, results_limited:)
...
       response_total
     end
   end
+
+  def is_done_page?(path)
+    path.start_with?("done", "/done")
+  end
+
+
+  def set_response_total(total_count, results_limited, path)
```
boffbowsh: This isn't "setting" a response total, `get_response_total` would be more suitable.

app/helpers/feedex_helper.rb
```diff
@@ -12,8 +11,22 @@ def total_responses_header(total_count:, from:, to:, results_limited:)
...
     else
       response_total
     end
   end
+
+  def is_done_page?(path)
+    path.start_with?("done", "/done")
+  end
+
+
```
boffbowsh: Extra line break here

app/helpers/feedex_helper.rb
mgrassotti: 👍

app/helpers/feedex_helper.rb
mgrassotti: 👍

PR #282, opened 2016-07-25
Support forms text changes

kalleth:
[Trello card](https://trello.com/c/2PxYo5tt/2-text-changes-to-support-forms)

Update the titles and descriptions of all support requests.

So that the class names continue to match, update the class names to reflect the new titles. This part i've kept in two commits, as I'm not entirely sure they're not just churn.

The more I work on this, the more I think there must be a cleaner way to do this. Seven classes (one for each type of request) each having its own Zendesk representation too seems overkill, but I don't want to tackle that kind of refactor yet.

The zendesk ticket classes seem the most likely targets for refactor because they only have two or three relatively static things that change for each one. I may be able to transform it into a YAML file with a "factory" or builder, but I'll mull that over before actually attempting it.

# Before
screenshot 2016-07-25 11 01 22

# After
screenshot 2016-07-25 11 01 08

- [ ] Check (with Al Massa) if any of the zendesk ticket tags need to change
- [ ] Deploy to integration for product acceptance/review.


lib/support/requests/accounts_permissions_and_training_request.rb
```diff
@@ -15,7 +15,7 @@ class CreateOrChangeUserRequest < Request
         end
       end
       validates_presence_of :action, :requested_user
-      validates :action, inclusion: { in: ["create_new_user", "change_user"] }
+      validates :action, inclusion: { in: ["create_new_user", "change_user", "unsuspend_user"] }
```
kalleth: I added `unsuspend_user` too because it was requested as a "pretty please" on the trello card.

app/controllers/requests_controller.rb
```diff
@@ -47,8 +46,4 @@ def set_requester_on(request)
     request.requester.name = current_user.name
     request.requester.email = current_user.email
   end
-
-  def persistence_to_zendesk_necessary?
```
NeilvB: Just a comment rather than an issue - we'll probably have to revisit this logic when we remove the zendesk dependency from anonymous feedback requests.

kalleth: meh.

I'd prefer to make them use a separate code path entirely at that stage rather than do magic metaprogramming.

NeilvB:
Looks good! 👍

kalleth:
I'm going to merge this now. Al (User Support Manager) has a list of the analytics tags that apply, and he can let us know if there's anything totally off with them.

It's not difficult to change these tags where necessary, so this PR can land as-is.

PR #285, opened 2016-07-26
Update content advice description

kalleth:
[Trello card](https://trello.com/c/2PxYo5tt/2-text-changes-to-support-forms)

> I have a very small amendment for one of the descriptions. Can this be amended? For content advice, need to add Request short URLs, topical event pages, groups or manuals. So the entire description would be: Ask for help or advice on any content problems. Request short URLs, topical event pages, groups or manuals.


NeilvB:
👍

PR #286, opened 2016-08-08
Add the ability to export global feedback

kalleth:
[Trello card](https://trello.com/c/ptXseJEJ/13-export-global-anonymous-feedback-csv)

screenshot 2016-08-08 15 28 53

This section of the support application (feedex) now allows the
total quantity of requests received each day to be exported as a CSV
file.

Initially, we're protecting this behind a new role (`feedex_exporters`)
so we can control who gets access to the functionality. The main
reasoning here is to allow side-by-side verification of the figures
before rolling it out to wider (perhaps 'single point of contact')
release.

### DIFF
--- a/app/controllers/anonymous_feedback/global_export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/global_export_requests_controller.rb
@@ -1,0 +1,20 @@
+require "gds_api/support_api"
+
+class AnonymousFeedback::GlobalExportRequestsController < AuthorisationController
+  def create
+    authorize! :request, :global_export_request
+
+    support_api.create_global_export_request(export_request_params)
+    redirect_to anonymous_feedback_explore_path,
+      notice: "We are sending your CSV file to #{current_user.email}. If you don't see it in a few minutes, check your spam folder."
+  end
+
+private
+  def export_request_params
+    params.permit(:from_date, :to_date).merge(notification_email: current_user.email)
+  end
+
+  def support_api
+    GdsApi::SupportApi.new(Plek.find("support-api"))
+  end
+end
--- a/app/views/anonymous_feedback/explore/new.html.erb
+++ b/app/views/anonymous_feedback/explore/new.html.erb
@@ -25,4 +25,29 @@
       <%= f.action :submit, label: "Explore by organisation", button_html: { class: "btn btn-success" } %>
     <% end %>
   </div>
+  <% if can? :request, :global_export_request %>
+    <div class="col-md-8">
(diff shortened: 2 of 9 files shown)


app/controllers/anonymous_feedback/global_export_requests_controller.rb
```diff
@@ -0,0 +1,20 @@
+require "gds_api/support_api"
+
+class AnonymousFeedback::GlobalExportRequestsController < AuthorisationController
+  def create
+    authorize! :request, :global_export_request
+
+    support_api.create_global_export_request(export_request_params)
+    redirect_to anonymous_feedback_explore_path,
+      notice: "We are sending your CSV file to #{current_user.email}. If you don't see it in a few minutes, check your spam folder."
```
elliotcm: Is this timeline ("a few minutes") realistic for long reporting periods?  If it's too short it could lead to an increase in zendesk tickets.

app/views/anonymous_feedback/explore/new.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <div class="row">
   <p class="lead col-md-8">
-    At the bottom of every page on GOV.UK we ask, “is there anything wrong with this page?”. Users can leave comments on what they were doing and what went wrong. Feedback is anonymous and we’ve not removed spam.
+  At the bottom of every page on GOV.UK we ask, “is there anything wrong with this page?”. Users can leave comments on what they were doing and what went wrong. Feedback is anonymous and we’ve not removed spam.
```
elliotcm: Why outdent this?

elliotcm:
Couple of minor questions otherwise 👍

app/controllers/anonymous_feedback/global_export_requests_controller.rb
kalleth: I just requested a report from 1st of January to the 8th of August 2016, and it took approximately 2 minutes to arrive.

This time window is far longer than the intended use case for this app so I'm comfortable sticking to the wording also used for the other exports here.

app/views/anonymous_feedback/explore/new.html.erb
kalleth: Typo

kalleth: or, more accurately, my vim autoindent had a special moment.

elliotcm:
👍 to merge when ready

kalleth:
I'm going to merge this now because Al has indicated that the data is awesomesauce.

PR #290, opened 2016-09-21
Mark problem reports as spam

NeilvB:
[Trello card](https://trello.com/c/g1I2TXF6/14-mark-anonymous-feedback-as-spam-download-filtered-report-xxl)

Expose the ability to review sets of problem reports in batches.

A user with the `feedex_explorers` permission is able to view pages of anonymous feedback and commit each page as a batch of reviewed feedback.

### Feedex landing page
screencapture-support-dev-gov-uk-anonymous_feedback-explore-1474471315967

### Review reports page
screencapture-support-dev-gov-uk-anonymous_feedback-problem_reports-1474470574523


config/routes.rb
```diff
@@ -11,12 +11,6 @@
   resources :named_contacts, only: :create
 
   namespace :anonymous_feedback do
-    get 'problem_reports', to: redirect {|p, req| req.params[:path] ? "/anonymous_feedback?path=" + req.params[:path] : "/anonymous_feedback"}
```
NeilvB: Kibana log search for this endpoint:
screen shot 2016-09-13 at 14 16 37

config/routes.rb
```diff
@@ -11,12 +11,6 @@
...
-
-    namespace :problem_reports do
-      get :explore, to: redirect("/anonymous_feedback/explore"), format: false
-    end
```
NeilvB: Kibana log search for this endpoint:
screen shot 2016-09-13 at 14 16 15

app/assets/stylesheets/application.scss
```diff
@@ -52,3 +52,25 @@ thead {
...
+  }
+
+  .mark-as-spam-checkbox label {
+    display: block;
+    padding: 12px;
+  }
+
+  .problem-report.reviewed,
+  .problem-report.reviewed .table-header-secondary {
+    background-color: #f4fef4;
```
kalleth: I know we're only using this in a single place, but can we set it up as a sass variable (`$light-green`) similar to how govuk_admin_template sets it up? Perhaps in `application.scss` -- see https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/stylesheets/govuk_admin_template/_theme.scss

app/controllers/anonymous_feedback/global_export_requests_controller.rb
```diff
@@ -11,7 +11,7 @@ def create
 
 private
   def export_request_params
-    params.permit(:from_date, :to_date).merge(notification_email: current_user.email)
+    params.permit(:from_date, :to_date).merge(notification_email: current_user.email, exclude_reviewed: !!params.permit(:exclude_reviewed))
```
kalleth: Nit: having two permits nested within each other / so close feels a bit suspect. Would this be cleaner as a helper which we then use?

```
def exclude_reviewed?
  !!params.permit(:exclude_reviewed)
end

params.permit(:from_date, :to_date).merge(notification_email: current_user.email, exclude_reviewed: exclude_reviewed?)
```

app/controllers/anonymous_feedback/problem_reports_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def fetch_problem_reports
+    AnonymousFeedbackApiResponse.new(
+      support_api.problem_reports(api_params).to_hash
+    )
+  end
+
+  def api_params
+    {
+      exclude_reviewed: !index_params[:include_reviewed],
```
kalleth: O_o

How come we have a disconnect here? I'm a little confused as to why we'd ask the user if they wanted to include reviewed or not and then invert that on the API side.

There's probably a reason, tho...

app/views/anonymous_feedback/explore/new.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <div class="row">
   <p class="lead col-md-8">
-    At the bottom of every page on GOV.UK we ask, “is there anything wrong with this page?”. Users can leave comments on what they were doing and what went wrong. Feedback is anonymous and we’ve not removed spam.
+    At the bottom of every page on GOV.UK we ask, “is there anything wrong with this page?”. Users can leave comments on what they were doing and what went wrong. <%= can?(:request, :review_feedback) ? 'Feedback is anonymous.' : 'Feedback is anonymous and we’ve not removed spam.' %>
```
kalleth: 👍

app/views/anonymous_feedback/problem_reports/_problem_report.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <%= link_to friendly_referrer(problem_report.referrer), problem_report.referrer, class: "breakable" %>
+      <% if search_term %>
+        <br /><span title="Search term">“<%= search_term %>”</span>
+      <% end %>
+    <% end %>
+  </td>
+  <td class="mark-as-spam-checkbox">
+    <%# Wrap checkbox in a label in attempt make targeting the box a little easier %>
+    <label for="<%= "mark_as_spam_#{problem_report.id}"%>">
+      <%= hidden_field_tag "mark_as_spam[#{problem_report.id}]", false, { id: '' } %>
```
kalleth: The commit message should probably contain some explanation for the hidden field duplication, as this isn't immediately obvious.

app/views/anonymous_feedback/problem_reports/index.html.erb
```diff
@@ -0,0 +1,88 @@
+<%= content_for :page_title, "Review Feedback" %>
+<%= content_for :header, "Review Feedback" %>
+
+<%= paginate @feedback, theme: 'twitter-bootstrap-3' %>
+
+<div>
+  <p>Reports rows highlighted in green have already been reviewed.</p>
```
kalleth: This is only true if viewing with "include reviewed" checked, correct? Is there a way we could reflect that in when we show this message?

config/routes.rb
kalleth: 👍

config/routes.rb
kalleth: 👍

config/routes.rb
```diff
@@ -11,19 +11,19 @@
...
     post :explore, to: "explore#create", format: false
 
     resources :organisations, only: :show, param: :slug, format: false
 
     resources :export_requests, only: [:create, :show], format: false
     resources :global_export_requests, only: [:create], format: false
+
+    resources :problem_reports, only: [:index] do
+      collection do
+        put :review
```
kalleth: If it's only this route, we can also use `put :review, on: :collection` to one-line it. Nit.

lib/support/permissions/ability.rb
```diff
@@ -18,6 +18,7 @@ def initialize(user)
 
         can :read, :anonymous_feedback
         can :request, :global_export_request if user.has_permission?('feedex_exporters')
+        can :request, :review_feedback if user.has_permission?('feedex_exporters')
```
kalleth: Yeah, this role is pretty bogusly named now. We should ask someone to add a new role to signon for the support app and use that instead. Perhaps `feedback_reviewers`.

spec/features/global_export_request_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+require 'gds_api/test_helpers/support_api'
+
+feature 'Exporting Global CSV of Feedback' do
```
kalleth: Did I forget to add a feature spec for this? Nice work :)

spec/features/review_anonymous_feedback_spec.rb
```diff
@@ -0,0 +1,408 @@
...
+feature 'Reviewing Anonymous Feedback' do
+  include GdsApi::TestHelpers::SupportApi
+
+  let(:user) { create :user_who_can_access_everything }
+
+  let(:unreviewed_problem_report_1) {
+    {
+      "id" => 1,
+      "type" => "problem-report",
+      "what_wrong" => "Yeti",
```
kalleth: They're known as Abominable Snowmen, i'll have you know! :D

app/controllers/anonymous_feedback/problem_reports_controller.rb
NeilvB: When I made the endpoint in the support api my thought process was that by default we would return everything (i.e. all reviewed and not reviewed), and then the various options filter that into subsets.

The front end is the other way around of course, so that only non-reviewed reports should be returned by default. Do you think it would be better for me to align this with the API?

app/views/anonymous_feedback/problem_reports/index.html.erb
NeilvB: Ah yes I could toggle this message to only display when that filter is applied 👍

PR #291, opened 2016-09-22
Support form titles

NeilvB:
[Trello card](https://trello.com/c/dyBzCD0H/33-change-titles-and-descriptions-inside-support-forms)

## UI copy changes

Support form titles are now the `label` of the request. The description is presented beneath using the blue "alert" box:

screen shot 2016-09-22 at 16 56 14

Navigation drop down has a new 'technical support' section:

screen shot 2016-09-22 at 16 56 27


kalleth:
Can we make sure this doesn't look weird with the two forms that have 'guidance' on them? (i.e. we're not displaying two blue ones one on top of another).

Otherwise, looks good :+1:

NeilvB:
Thanks @kalleth I hear what you're saying - that was my interpretation of the ticket and thought I'd go ahead with implementation as it's a quick one. Will see what the product owner says.

PR #292, opened 2016-09-30
Add worker spec for 302 response from Zendesk

NeilvB:
[Trello card](https://trello.com/c/GgAfSAFK/42-harden-the-zendesk-integration-and-improve-monitoring)


kalleth:
build failure, otherwise looks good :) :+1: :shipit: -- I assume the build failure is because the helper isn't available yet because the GdsZendesk gem hasn't been updated per the other PR?

PR #294, opened 2016-11-08
Allow all users to view emergency contact details

boffbowsh:


NeilvB:
This looks good

tijmenb:
Can you make this user a normal user? Then this would be tested.

https://github.com/alphagov/support/blob/master/spec/features/emergency_contact_details_spec.rb#L9

boffbowsh:
Good shout @tijmenb

tijmenb:
👍

PR #295, opened 2016-11-09
Permit all users to make analytics requests

NeilvB:
A previous change (from [card](https://trello.com/c/COcqtgck/29-modify-support-permissions)) removed the ability to access this feature from all users and instead only permitted those with 'single point of access' and 'user managers' permissions. This commit reinstates the ability of all users to make analytics requests following feedback from the support team.


h-lame:
:+1:

PR #299, opened 2017-01-13
Add Jenkinsfile

mgrassotti:
Trello: https://trello.com/c/TXb7NOrK/570-move-support-to-new-ci-infrastructure-1

The old CI infrastructure is in the process of being wound down so we
need to move our apps to the new one.


Jenkinsfile
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'support'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
+node('mongodb-2.4') {
```
steventux: https://ci.dev.publishing.service.gov.uk/job/govuk_support/configure doesn't restrict where this project can be built so we don't need to restrict it to a mongodb-2.4 node here. Should be:
```
node {
...
}
```

mgrassotti: 👍

PR #312, opened 2017-05-11
Get application files to Rails 4.2 standard

h-lame:
I ran `rake rails:update` to update existing files and generate new ones. This means our app is now configured like a Rails 4.2 app should be, and this will make it easier to upgrade to Rails 5.

This commit also migrates the app to use `secrets.yml`, and importantly this means it uses the `SECRET_KEY_BASE` env var on deployed environments. The infrastructure work to support this has already been done so we can deploy this without further work. Prior to this we relied on the
`secret_token.rb` initializer being overwritten by our deployment process, but that functionality was stopped in Dec 2016, meaning the app has been insecure since then.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,0 +1,29 @@
+#!/usr/bin/env ruby
+require 'pathname'
+
+# path to your application root.
+APP_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
+
+Dir.chdir APP_ROOT do
+  # This script is a starting point to setup your application.
+  # Add necessary setup steps to this file:
+
+  puts "== Installing dependencies =="
+  system "gem install bundler --conservative"
+  system "bundle check || bundle install"
+
+  # puts "\n== Copying sample files =="
+  # unless File.exist?("config/database.yml")
+  #   system "cp config/database.yml.sample config/database.yml"
+  # end
+
+  # puts "\n== Preparing database =="
+  # system "bin/rake db:setup"
+
+  puts "\n== Removing old logs and tempfiles =="
+  system "rm -f log/*"
+  system "rm -rf tmp/cache"
+
+  puts "\n== Restarting application server =="
+  system "touch tmp/restart.txt"
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,36 +1,20 @@
 require File.expand_path('../boot', __FILE__)
 
-# Pick the frameworks you want:
-# require "active_record/railtie"
+# require 'rails/all'
 require "action_controller/railtie"
 require "action_mailer/railtie"
 require "sprockets/railtie"
 require "rails/test_unit/railtie"
 
-require 'yaml'
-
-if defined?(Bundler)
-  # If you precompile assets before deploying to production, use this line
-  Bundler.require(*Rails.groups(:assets => %w(development test)))
-  # If you want your assets lazily compiled in production, use this line
-  # Bundler.require(:default, :assets, Rails.env)
-end
+# Require the gems listed in Gemfile, including any gems
+# you've limited to :test, :development, or :production.
+Bundler.require(*Rails.groups)
 
 module Support
   class Application < Rails::Application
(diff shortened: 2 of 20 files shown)


config/application.rb
```diff
@@ -1,73 +1,30 @@
...
-    # Enable the asset pipeline
-    config.assets.enabled = true
-
-    # Version of your assets, change this if you want to expire all your assets
-    config.assets.version = '1.0'
-    config.assets.prefix = '/assets'
 
-    config.assets.precompile += %W(html5.js)
+    # Do not swallow errors in after_commit/after_rollback callbacks.
+    config.active_record.raise_in_transactional_callbacks = true
```
gpeng: I think this is causing the test failure because `active_record` isn't loaded.

config/locales/en.yml
```diff
@@ -0,0 +1,23 @@
...
+# To use a different locale, set it with `I18n.locale`:
+#
+#     I18n.locale = :es
+#
+# This would use the information in config/locales/es.yml.
+#
+# To learn more, please read the Rails Internationalization guide
+# available at http://guides.rubyonrails.org/i18n.html.
+
+en:
```
gpeng: Do we need this? In fact, do we need this file?

h-lame: Do we need it; probably not.  Is it worth having so the diff for future rails upgrades is minimal and we don't need to worry about it again; I'd say so, yes.

config/application.rb
h-lame: d'oh!  Thanks @gpeng !

bin/setup
```diff
@@ -0,0 +1,29 @@
...
+  puts "== Installing dependencies =="
+  system "gem install bundler --conservative"
+  system "bundle check || bundle install"
+
+  # puts "\n== Copying sample files =="
+  # unless File.exist?("config/database.yml")
+  #   system "cp config/database.yml.sample config/database.yml"
+  # end
+
+  puts "\n== Preparing database =="
```
andrewgarner: We should comment out these steps as we're not using ActiveRecord.

config/environments/development.rb
```diff
@@ -1,30 +1,41 @@
...
   config.action_mailer.raise_delivery_errors = false
 
-  # Print deprecation notices to the Rails logger
+  # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
 
-  # Only use best-standards-support built into browsers
-  config.action_dispatch.best_standards_support = :builtin
+  # Raise an error on page load if there are pending migrations.
+  config.active_record.migration_error = :page_load
```
andrewgarner: We'll need to comment this out.

config/environments/production.rb
```diff
@@ -1,67 +1,80 @@
...
 
-  # Send deprecation notices to registered listeners
+  # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Use default logging formatter so that PID and timestamp are not suppressed.
+  config.log_formatter = ::Logger::Formatter.new
+
+  # Do not dump schema after migrations.
+  config.active_record.dump_schema_after_migration = false
```
andrewgarner: And this.

config/locales/en.yml
gpeng: Yep... fair enough.

andrewgarner: We don't need it, but the comments at the top of the file are useful to anyone who might look to add future translations. It also makes the upgrade process to Rails 5 easier as out application will be closer to that off a scaffolded application and thus the diff should be simpler.

config/application.rb
h-lame: commented out all active_record config options and rebased.  I figure that leaving them in, but commented out, is best so the diff for future upgrades is easier to understand.

bin/setup
h-lame: Oh, yes.  Will do.

h-lame: Done and rebased.

PR #314, opened 2017-05-24
Refactors models 

elenatanasoiu:
This PR does three things: 

1. Moves models from lib to app/models. 
2. Removes redundant requires and includes, but also takes advantage of automatic loading of models in the Rails initialization process.
3. As a result of step 2, we now use full namespacing to refer to objects

[Trello card](https://trello.com/c/lgoeQX76/122-investigate-refactoring-support-project)


app/models/zendesk/zendesk_tickets.rb
```diff
@@ -1,14 +1,12 @@
...
-        :collaborators => ticket_to_raise.collaborator_emails,
-        :tags => ticket_to_raise.tags,
-        :comment => { "body" => ticket_to_raise.comment }
-      })
+      ZendeskTicketWorker.perform_async(subject: ticket_to_raise.subject,
+        priority: ticket_to_raise.priority,
+        requester: { "locale_id" => 1, "email" => ticket_to_raise.email, "name" => ticket_to_raise.name },
+        collaborators: ticket_to_raise.collaborator_emails,
+        tags: ticket_to_raise.tags,
+        comment: { "body" => ticket_to_raise.comment })
```
h-lame: This looks pretty weird.  Would it be happy with:

```ruby
ZendeskTicketWorker.perform_async(
  subject: ticket_to_raise.subject,
  priority: ticket_to_raise.priority,
  requester: { "locale_id" => 1, "email" => ticket_to_raise.email, "name" => ticket_to_raise.name },
  collaborators: ticket_to_raise.collaborator_emails,
  tags: ticket_to_raise.tags,
  comment: { "body" => ticket_to_raise.comment })
)
```

instead? (E.g. keep the new hash syntax and drop the redundant `{}` but don't move the first part of the hash onto the same line as the method call.)

app/models/zendesk/zendesk_ticket.rb
```diff
@@ -80,7 +80,7 @@ def base_attribute_snippets
 
     def has_value?(param, target = nil)
       target ||= @request
-      target.respond_to?(param) and not target.send(param).blank?
+      target.respond_to?(param) && !target.send(param).blank?
```
h-lame: I think we can squash this commit down with the automatic lint commit - there's no real need to distinguish them, particularly as these manual changes look like the kind of thing that could have been automatic.

app/models/support/gds/user_facing_components.rb
```diff
@@ -1,5 +1,3 @@
-require 'support/gds/user_facing_component'
```
h-lame: We've got future commits that remove the redundant requires, so for consistency we should move this change (and the others in this commit that aren't just about moving the files) into those commits.

app/models/zendesk/zendesk_ticket.rb
elenatanasoiu: Done

app/models/support/gds/user_facing_components.rb
elenatanasoiu: Done

app/models/zendesk/zendesk_tickets.rb
elenatanasoiu: The linter approves of this.

app/models/zendesk/ticket/content_change_request_ticket.rb
```diff
@@ -30,3 +30,4 @@ def comment_snippets
     end
   end
 end
+u
```
h-lame: This probably shouldn't be here.

PR #315, opened 2017-05-30
[DO NOT MERGE] Require 'signin' permission from signon to access app

h-lame:
For: https://trello.com/c/FMBZdTvk/45-support-app-should-probably-require-signon-permission-to-access

The expected behaviour of apps using signon is that they authenticate the
user and check that they have the 'signin' permission before allowing the
user to use the app.  This app only did the first part, not the second
bit.  The upside is that all signon users could access the support app;
the downside is that we have no way to limit a user so they can't use
this app, but still use other apps.  In practice this may not really be a
problem because of the purpose of the support app - it's unlikely we'd
want to allow someone to access other apps, but not support.  However it
does mean that the support app acts differently with respect to the
'signin' permission that is managed in signon and so this could be
confusing to future developers.

We will do some data work in signon to make sure that all existing users
retain signin permission, and that all new users are granted signin
permission for support by default.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
 
   before_action :authenticate_support_user!
+  before_action :require_signin_permission!
 
   protect_from_forgery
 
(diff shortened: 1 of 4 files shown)


h-lame:
Flagged as DNM because we need to do the data work in signon before we deploy this change.

h-lame:
Closing this for now, in preference to implementing the recommendations in [RFC 78](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-078-re-architect-signin-permissions-in-signon.md), which we're tracking in [this trello card](https://trello.com/c/c4bgclDp/171-implement-rfc-78-make-signon-handle-signin-permission-during-oauth)

PR #318, opened 2017-09-01
Add ui for filtering by path and org to feedex

h-lame:
For: https://trello.com/c/8wwFr5y0/196-improvements-to-feedex-ui

It's possible to filter feedex by both path and organisation if you manipulate the URL, but there's no UI so it's not obvious.  This PR adds UI elements to make it more obvious and easier to do.

## Screenshots

We choose to browse by org, and then decide we want to also filter by a specific path, and finally remove the org constraint.

### Before

feedex-before

There's no UI for adding path or removing organisation filters, but we can do it by manually hacking the URL.  The page doesn't tell us we've filtered by multiple things, but the data has been updated accordingly.

### After

feedex-after

UI for path and org filtering is always present on the page, and we've tweaked the help text to make it clear you can use a partial path to filter results.

##### V1 - icon button in header

feedex-after-v1

##### V2 - text button in form

feedex-after-v2

---

Would particularly appreciate thoughts on the UI choices I've made as I'm not 100% sure they're the best.

### DIFF
--- a/app/presenters/scope_filters_presenter.rb
+++ b/app/presenters/scope_filters_presenter.rb
@@ -4,7 +4,7 @@ class ScopeFiltersPresenter
   attr_reader :path, :organisation_slug
 
   def initialize(path: nil, organisation_slug: nil)
-    @path = path
+    @path = normalize_path(path)
     @organisation_slug = organisation_slug
   end
 
@@ -40,6 +40,19 @@ class ScopeFiltersPresenter
   end
 
 private
+
+  def normalize_path(path_or_url)
+    return nil unless path_or_url.present?
+    normalized_path = URI.parse(path_or_url).path
+    if normalized_path.present?
+      normalized_path.sub!(/^(http(s)?(:)?(\/)+?(:)?)?((\/)?www.)?gov.uk/, '')
+      normalized_path.start_with?('/') ? normalized_path : "/#{normalized_path}"
+    else
+      '/'
+    end
+  rescue URI::InvalidURIError
+    path_or_url
+  end
 
   def support_api
     GdsApi::SupportApi.new(Plek.find("support-api"))
--- a/app/presenters/scope_filters_presenter.rb
+++ b/app/presenters/scope_filters_presenter.rb
@@ -1,0 +1,47 @@
+require 'gds_api/support_api'
+
+class ScopeFiltersPresenter
+  attr_reader :path, :organisation_slug
+
+  def initialize(path: nil, organisation_slug: nil)
+    @path = path
+    @organisation_slug = organisation_slug
+  end
+
+  def filtered?
+    path.present? || organisation_slug.present?
+  end
+
+  def invalid_filter?
+    !filtered?
+  end
+
+  def done_page?
+    path.present? ? path.start_with?("done", "/done") : false
+  end
+
+  def organisation_title
+    organisation["title"] if organisation.present?
+  end
+
+  def organisation
+    @organisation ||= support_api.organisation(organisation_slug) if organisation_slug.present?
+  end
+
+  def to_s
+    if invalid_filter?
+      "Everything"
+    else
+      [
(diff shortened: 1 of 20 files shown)


fofr:
Can we avoid magic icons and have a link with text? eg "Filter by path", "Edit filters", or something?

h-lame:
@fofr - how does the v2 gif look (I've updated the description above).

h-lame:
@fofr / other interested parties: I reviewed this with Liz and we decided that hiding the UI was counter-intuitive even though having it there all the time pushes the results further down the page.  We also clarified the help text for URL filters to make it clearer that you can use a partial path while searching to find feedback on entire classes of documents (like guidance or done pages).

spec/features/feedex_spec.rb
```diff
@@ -103,4 +104,164 @@
...
+            type: "problem-report",
+            path: "/guidance/doing-bad-things",
+            url: "http://www.dev.gov.uk/guidance/doing-bad-things",
+            created_at: DateTime.parse("2013-02-01"),
+            what_doing: "looking at what you consider bad",
+            what_wrong: "so many typos, whoever wrote this must have been angry",
+            referrer: "https://www.gov.uk/having-fun",
+          },
+        ]
+      }
```
elenatanasoiu: Can't this hash re-use the one used here: https://github.com/alphagov/support/pull/318/commits/01451a9898ceb7fc8d2fea9a9467c39815139ae1#diff-2831018522e781bb592490f9d762a4a6R122

As far as I can tell the thing that differs between these requests is the params in the first hash, so the second hash could probably be defined once and then re-used everywhere.

app/presenters/scope_filters_presenter.rb
```diff
@@ -1,13 +1,17 @@
...
+  attr_reader :organisation_slug
 
   def initialize(path: nil, organisation_slug: nil)
-    @path = path
+    @requested_path = path
     @organisation_slug = organisation_slug
   end
 
+  def path
+    @path ||= path_from_requested_path if requested_path.present?
```
elenatanasoiu: Can `path` be renamed to `parsed_path` so we can tell from the start that it's the result of doing things to the requested path? 

There might be some confusion about what the difference is between `path` and `requested_path`, especially since it shares a name with a parameter sent to the `initialize` method above. And that parameter refers to the original path, before any parsing.

spec/features/feedex_spec.rb
```diff
@@ -107,122 +105,117 @@
...
-            what_wrong: "I know they're not really a hat, but would it hurt to mention them",
-            referrer: "https://www.gov.uk/hats",
-          },
-        ]
-      }
+      "current_page" => 1,
+      "pages" => 1,
+      "page_size" => 2,
+      "results" => [
+        {
```
elenatanasoiu: Same as before, I think this might be shortened by assigning this has once and then re-using the it in multiple scenarios, since the only that really needs to vary afaik is the value of the params received in the first hash (e.g. `{ path_prefix: '/guidance' }` )

h-lame: You're right that the only thing that needs to vary is the args in the query hash, and the actual results don't matter. However, I am a fan of making sure that test data is realistic (because we never know how the app might change in the future and our tests break if the data isn't correct).  You're right that it's quite noisy though, I'll pull out these stubs to methods and see if I can make the test shorter.

h-lame: Did this in fb09e4378cd7233332333cf50f253abbcbe5ff89

app/presenters/scope_filters_presenter.rb
h-lame: Good call, did a0b7e44c1159e63dd0b5e061351cfa8f9fc79e5a as a fixup commit which I'll squash down if you think it's clearer (or I will take another crack at if you don't)

PR #321, opened 2017-09-28
Copy changes to emergency contacts page

keithiopia:
I've simplified the names of the options - it's more obvious which number to call at which time of day now. Also prioritised display of the phone numbers.

@jonsanger has checked the copy.

Before:
emergency contact page old

After:
emergency contact screenshot


jonsanger:
Looks good

PR #323, opened 2017-10-24
Fix Zendesk name too long record invalid error

theKHutDeveloper:
A Zendesk ticket is created as long as the requester name is less than 256 characters.
This will remove the Zendesk name to long errors in Sentry.

We (2nd line) started looking at this issue as there were a large number of these errors in Sentry, when we looked at these errors they appears to be spam. We spoke with Liz Lutgendorff and she suggested ignoring any requests there the name was above the 255 limit as these were always spam and not requests from 'real' users.

### DIFF
--- a/app/workers/zendesk_ticket_worker.rb
+++ b/app/workers/zendesk_ticket_worker.rb
@@ -6,7 +6,11 @@ class ZendeskTicketWorker
       GovukStatsd.increment("report_a_problem.submission_from_suspended_user")
     else
       begin
-        create_ticket(ticket_options)
+        if ticket_options["requester"]["name"] && ticket_options["requester"]["name"].length > 255
+          GovukStatsd.increment("report_a_problem.zendesk_ticket_name_too_long")
+        else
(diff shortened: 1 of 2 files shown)


suzannehamilton:
Here's the Trello card for this issue: https://trello.com/c/btVExJP7/379-identify-and-fix-high-volume-sentry-errors

app/workers/zendesk_ticket_worker.rb
```diff
@@ -6,7 +6,11 @@ def perform(ticket_options)
       GovukStatsd.increment("report_a_problem.submission_from_suspended_user")
     else
       begin
-        create_ticket(ticket_options)
+        if ticket_options["requester"]["name"] && ticket_options["requester"]["name"].length > 255
```
cbaines: The try and dig methods could be used here, but I'll leave it up to you to decide if this is a good idea or not, e.g.
```ruby
if ticket_options.dig("requester", "name").try { |name| name.length > 255 }
```

PR #325, opened 2017-11-15
Improve README

elenatanasoiu:
For: https://trello.com/c/HeSzjWE8/293-improve-readme-for-support-application

Makes the README compliant with the [GOV.UK README styleguide](https://docs.publishing.service.gov.uk/manual/readmes.html).

More information has been added about the purpose of the application. We've made an effort to keep the README scannable for a first time user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,71 @@
 # Support
 
-Forms that create Zendesk tickets for requests coming from government agencies.
-The app is written in Rails and uses the zendesk_api gem to connect to Zendesk.
+Forms that create tickets for requests coming from government agencies or from GOV.UK contact forms.
 
-### Getting started
+## Nomenclature
+
+- **Feedback**: All the data received from contact forms is considered to be "feedback" 
+of some form or other and relates to pages published on GOV.UK. 
+- **Anonymous Contact**: Part of the feedback collected by this app is anonymous, when it's 
+submitted via an anonymous contact form in the [feedback app][feedback]. 
+- **Named Contact**: In contrast with the `Anonymous Contact` feedback, this is submitted 
+via a form that will require you to identify yourself. This data is sent directly 
+to the `support` app. 
+
+## Technical documentation
+
+This is a Ruby on Rails application which collects feedback and creates tickets based on it. 
+It collects data from two sources: 
+- its own forms which can be accessed by [signon][signon] users
+- contact forms from the [feedback app][feedback]
+
+The data collected from either the `feedback` app or its own forms is then used to create 
+and submit Zendesk tickets.
+
+Even though it is not public facing, the `support` app does receive data from the [feedback app][feedback], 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,25 @@
 # Support
 
 Forms that create Zendesk tickets for requests coming from government agencies.
-The app is written in Rails and uses the zendesk_api gem to connect to Zendesk.
 
-### Getting started
+## Technical documentation
+
+This is a Ruby on Rails application which is not public facing. However, it does face GOV.UK employees and is used as a way to submit Zendesk tickets.
```
h-lame: GOV.UK employees isn't quite right.  It's available to all signon users, most of whom don't even work for GDS, let alone work on GOV.UK.  I think it's being opened up to local government shortly too.

README.md
```diff
@@ -1,9 +1,25 @@
...
 
-### Getting started
+## Technical documentation
+
+This is a Ruby on Rails application which is not public facing. However, it does face GOV.UK employees and is used as a way to submit Zendesk tickets. 
+The Zendesk tickets are created either from the information submitted in the `support` app forms or via the information collected from public 
+users which is forwarded from the `feedback` app. The `feedback` app is public facing and serves its own contact forms.
+
+The app makes use of the `gds_zendesk` gem to connect to Zendesk. 
+
```
h-lame: We should mention the feedback explorer which allows users to browse the feedback submitted from GOV.UK pages, and to export CSVs of the same for further analysis.

I think it might also be worth de-emphasising zendesk as it raises the question "what's this app for? why not just use zendesk directly?".  The support app lets people who use the GOV.UK publishing tools ask for specific help (creating a new campaign site, making content changes, asking for user permissions, asking for feature requests, etc...) from the GOV.UK team.   It may be worth listing all these types, or pointing at the place in the code where these request types are defined (if it's easy to interpret).

README.md
```diff
@@ -1,9 +1,25 @@
...
 
 Forms that create Zendesk tickets for requests coming from government agencies.
-The app is written in Rails and uses the zendesk_api gem to connect to Zendesk.
 
-### Getting started
+## Technical documentation
+
+This is a Ruby on Rails application which is not public facing. However, it does face GOV.UK employees and is used as a way to submit Zendesk tickets. 
+The Zendesk tickets are created either from the information submitted in the `support` app forms or via the information collected from public 
+users which is forwarded from the `feedback` app. The `feedback` app is public facing and serves its own contact forms.
```
h-lame: Maybe add a link to the feedback app?  It's not a dependency exactly, because it's really something that sits upstream of the support app, so it won't appear below in the dependencies section.

elenatanasoiu:
I've added a "Nomenclature" section

h-lame:
Good stuff!

PR #328, opened 2017-12-07
Explore feedback by multiple paths or by format

elenatanasoiu:
For: https://trello.com/c/IB9P5EyE/266-introduce-new-browse-options-for-feedex

This will allow you to search:
- by multiple paths at the same time by writing them in the "By URL" search box. For ex: `/vat-rates, /guidance/, https://www.gov.uk/bank-holidays`
- by multiple paths at the same time by uploading a list of comma-separated paths in the  "Upload a URL List" section. Example contents of file: `/vat-rates, /guidance/, https://www.gov.uk/bank-holidays`
- by page document type in the "By document type" section. Example for document type: `Smart Answer`

screen shot 2017-12-13 at 17 13 25
screen shot 2017-12-13 at 17 13 35
screen shot 2017-12-14 at 12 16 39


app/controllers/anonymous_feedback/explore_controller.rb
```diff
@@ -9,13 +9,21 @@ def new
     @organisations_list = support_api.organisations_list.map do |org|
       [organisation_title(org), org["slug"]]
     end
+    @explore_by_document_type = Support::Requests::Anonymous::ExploreByDocumentType.new
+    @document_type_list = support_api.document_type_list.to_a.map do |document_type|
+      [document_type.titleize, document_type]
+    end
```
h-lame: You're just following the established pattern for orgs, but I wonder if it would be a good refactor to move this .to_a.map into a helper or something because we use it in two places and it'd be a shame if one got out of sync.  And to do the same for the org list too.

app/controllers/anonymous_feedback/explore_controller.rb
```diff
@@ -9,13 +9,21 @@ def new
     @organisations_list = support_api.organisations_list.map do |org|
       [organisation_title(org), org["slug"]]
     end
+    @explore_by_document_type = Support::Requests::Anonymous::ExploreByDocumentType.new
+    @document_type_list = support_api.document_type_list.to_a.map do |document_type|
```
h-lame: Out of interest, why do we need `.to_a` here, but not on the org list above?

app/views/anonymous_feedback/explore/new.html.erb
```diff
@@ -12,19 +12,41 @@
...
+    <h2 class="add-bottom-margin">Upload a URL List</h2>
+    <%= semantic_form_for @explore_by_multiple_paths, url: { action: "create" }, html: { class: 'well', multipart: true } do |f| %>
+      <%= f.inputs do %>
+        <%= f.input :uploaded_list, as: :file, label: "Upload list of URLs", required: true, input_html: { class: "input-md-7", :"aria-required" => true, type: "file"} %>
+      <% end %>
+      <p class="help-block add-bottom-margin">
+        The contents of the file should be a comma separated list:
+        <code>/vat-rates, https://www.gov.uk/bank-holidays, /guidance/</code>
+      </p>
+      <%= f.action :submit, label: "Upload list of urls", data: { confirm: "This will email you a CSV export. Are you sure you want to continue?" }, button_html: { class: "btn btn-success" } %>
```
h-lame: Doesn't this go to the browse page now? We can drop the "This will emails you a CSV export" warning.

app/presenters/scope_filters_presenter.rb
```diff
@@ -1,29 +1,42 @@
...
+    paths.present? || organisation_slug.present?
   end
 
   def invalid_filter?
     !filtered?
   end
 
   def done_page?
-    path.present? ? path.start_with?("done", "/done") : false
+    paths.present? ? paths.split(', ').any? { |path| path.start_with?("done", "/done") } : false
```
h-lame: We do `paths.split(', ')` a few times, maybe we should use `paths_for_api` because it's already an array?  We might want to think up a better name for that though, because `paths_for_api` makes sense in the place we use it outside the class, but not really inside.... 🤔

app/presenters/scope_filters_presenter.rb
```diff
@@ -1,29 +1,42 @@
 require 'gds_api/support_api'
 
 class ScopeFiltersPresenter
-  attr_reader :path, :organisation_slug
+  attr_reader :paths, :organisation_slug, :paths_for_api
 
-  def initialize(path: nil, organisation_slug: nil)
-    @path = normalize_path(path)
+  def initialize(paths: nil, organisation_slug: nil)
+    @paths = normalize_paths(paths)
```
h-lame: I find it sort of surprising that `normalize_paths` returns a string joined by commas, a return value that we store on `@paths`, but when you look into the implementation `normalize_paths` has also sets `@paths_for_api` to the array version of that string.  I think it would be clearer if either both assignments happened inside the method, or both happened outside the method.

h-lame: Following on from my comment on `done_page?` maybe we need something like:

```ruby
def initialize(paths: nil, ...)
  @some_internal_var_name_for_the_array_of_paths = normalize_paths(paths)
  ...
end

def paths
  some_internal_var_name_for_the_array_of_paths.join(', ')
end

def paths_for_api
  some_internal_var_name_for_the_array_of_paths
end

private
attr_reader :some_internal_var_name_for_the_array_of_paths
```

?

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -1,15 +1,20 @@
...
     end
 
     it "extracts the path from an http URL" do
-      presenter = described_class.new(path: 'http://www.gov.uk/some-path')
-      expect(presenter.path).to eq("/some-path")
+      presenter = described_class.new(paths: ['http://www.gov.uk/some-path', 'http://www.gov.uk/other-path'])
+      expect(presenter.paths).to eq('/some-path, /other-path')
+    end
+
+    it "extracts multiple paths separated by commas" do
```
h-lame: This test name suggests to me that some parsing of a string value for paths is going to happen, but we're providing an array with two elements.  The previous two tests also provide a multi-element array so I'm not sure what this test is actually trying to do - perhaps it's a hangover from an earlier version of the implementation and it's no longer required?

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -50,90 +55,126 @@
...
+      presenter = described_class.new(paths: ['/done/buying-a-new-hat'], organisation_slug: 'department-of-hats')
       expect(presenter).to be_filtered
     end
   end
 
-  describe "#invalid_filter?" do
-    it 'is true when path and organisation are blank' do
-      presenter = described_class.new(path: nil, organisation_slug: nil)
+  describe '#invalid_filter?' do
+    it 'is true when path`` and `organisation_slug` are blank' do
```
h-lame: Minor: the backticks here should be around `path`

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -50,90 +55,126 @@
...
+    context "with multiple paths" do
+      it 'is true when one of the provided paths starts with "/done"' do
+        presenter = described_class.new(paths: ['vat-rates', '/guidance/', '/done/buying-a-new-hat'])
+        expect(presenter).to be_done_page
+      end
+
+      it 'is true when one of the provided paths starts with "done"' do
+        presenter = described_class.new(paths: ['vat-rates', '/guidance/', '/done/buying-a-new-hat'])
+        expect(presenter).to be_done_page
+      end
```
h-lame: Minor: looks like this test is a duplicate of the preceding one?

spec/controllers/anonymous_feedback/export_requests_controller_spec.rb
```diff
@@ -18,34 +18,33 @@
...
         stub_support_api_feedback_export_request_creation(notification_email: "foo.bar@example.gov.uk",
-                                                      path_prefix: "/foo",
+                                                      path_prefixes: ["/foo"],
                                                       from: "2015-05-01",
                                                       to: "2015-06-01",
                                                       organisation: nil)
       end
 
-      let(:do_request) { post :create, params: { path: "/foo", from: "2015-05-01", to: "2015-06-01" } }
+      let(:do_request) { post :create, params: { paths: ["/foo"], from: "2015-05-01", to: "2015-06-01" } }
```
h-lame: In the views we set a `paths` hidden field set to the value of `ScopeFilterPresenter#paths` which is a comma separated string, but here we're testing using a `paths` param that is an array of strings.  I've just spotted the `.instance_of?(String)` in the controller itself and I think it's only there to get around this test.  If we use a string here then we can drop that type check.

app/controllers/anonymous_feedback/explore_controller.rb
```diff
@@ -1,21 +1,27 @@
...
 
   def new
     @explore_by_multiple_paths = Support::Requests::Anonymous::ExploreByMultiplePaths.new
     @explore_by_organisation = Support::Requests::Anonymous::ExploreByOrganisation.new(organisation: current_user.organisation_slug)
-    @organisations_list = support_api.organisations_list.map do |org|
-      [organisation_title(org), org["slug"]]
-    end
+    @organisations_list = parse_organisations(support_api.organisations_list)
+    @explore_by_document_type = Support::Requests::Anonymous::ExploreByDocumentType.new
+    @document_type_list = parse_doctypes(support_api.document_type_list)
```
h-lame: :+1:

spec/controllers/anonymous_feedback/document_types_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    stub_support_api_anonymous_feedback_doc_type_summary(
+      "smart_answer",
+      "last_7_days",
+      smart_answer_last_7_days_summary_results
+    )
+    login_as create(:user, organisation_slug: "cabinet-office")
+  end
+
+  describe "#show" do
+    it "makes a successful create request" do
```
h-lame: minor: "create request" doesn't seem like the right description for this test

spec/controllers/anonymous_feedback/explore_controller_spec.rb
```diff
@@ -29,6 +31,18 @@
     expect(assigns(:explore_by_multiple_paths)).to be_present
   end
 
+  it "shows the new form again for invalid organisation requests" do
+    post :create, params: { support_requests_anonymous_explore_by_organisation: { organisation: "" } }
+    expect(response).to have_http_status(422)
+    expect(assigns(:explore_by_document_type)).to be_present
+  end
```
h-lame: :+1: for adding missing tests for things you haven't changed!

spec/presenters/document_type_summary_presenter_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    let(:path_a) { "/path-a" }
+    let(:path_b) { "/path-b" }
+    let(:path_c) { "/path-c" }
+    let(:api_response) {
+      {
+          "document_type" => "smart_answer",
+          "anonymous_feedback_counts" => [
+              { "path" => path_c },
+              { "path" => path_a },
+              { "path" => path_b },
```
h-lame: Minor: This is probably rubocop being weird, but the indentation here is 4 instead of 2 spaces.

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -33,11 +34,20 @@ def export_request_params
   end
 
   def anonymous_feedback_params
-    @anonymous_feedback_params ||= params.permit(:from, :to, :path, :organisation).to_h
+    clean_paths
+    @anonymous_feedback_params ||= params.permit(:from, :to, :organisation, :path, paths: []).to_h
```
h-lame: I think we always send `paths` as a comma separated string now so we should be permitting paths as a scalar (`:paths`) not as an array (`paths: []`).  I'm surprised a test hasn't failed to expose this as I think without this the controller would see a param like `paths=/foo,/bar` as `nil`.

app/presenters/scope_filters_presenter.rb
```diff
@@ -1,29 +1,42 @@
...
+    paths.present? ? paths.split(', ').any? { |path| path.start_with?("done", "/done") } : false
   end
 
   def organisation_title
     organisation["title"] if organisation.present?
   end
 
+  def paths_title
+    if paths.present?
+      parsed_paths = paths.split(',').map(&:strip)
```
h-lame: Oh! Maybe `parsed_paths` is a good internal name for the array version of `paths` / `paths_for_api`?

spec/controllers/anonymous_feedback_controller_spec.rb
```diff
@@ -29,11 +29,11 @@
     context "on the first page" do
       it "should show no results" do
         stub_support_api_anonymous_feedback(
-          { path_prefix: "/a" },
+          { path_prefixes: ["/a"] },
           "results" => [], "pages" => 0, "current_page" => 1,
         )
 
-        get :index, params: { "path" => "/a", "format" => "json" }
+        get :index, params: { "paths" => ["/a"], "format" => "json" }
```
h-lame: I think throughout this test we should use `paths: "<comma separated string>"` instead of `paths: [<array of strings>]` because the UI will generate the former, not the latter.

app/controllers/anonymous_feedback/export_requests_controller.rb
elenatanasoiu: `clean_paths` will split `paths` into an array, before going through the param filtering. So `paths` is guaranteed to be an array every time.

h-lame: You're right.  My bad.

spec/controllers/anonymous_feedback/export_requests_controller_spec.rb
elenatanasoiu: I've added two extra tests to illustrate that the `paths` param can be 
- an array of values: ["/foo", "/bar", "/baz] (test already existed)
- a string with a comma separated list of values: "/foo, /bar, /baz" (new test)
- a string with a single value: "/foo" (new test)

Because `clean_paths` runs before the params.permit method in the controller, it will take all these three possible scenarios and always return `paths` as an array of values.

spec/presenters/scope_filters_presenter_spec.rb
elenatanasoiu: The preceding test checks for paths containing "/done", while this one checks for paths containing "done" without the forward slash. It's maybe a bit too thorough.

app/presenters/scope_filters_presenter.rb
elenatanasoiu: I've used `parsed_path` in place of `:some_internal_var_name_for_the_array_of_paths`.

spec/presenters/scope_filters_presenter_spec.rb
h-lame: So it does, I clearly wasn't paying enough attention.  It's good to be thorough!

app/presenters/scope_filters_presenter.rb
elenatanasoiu: `parsed_for_api` is now only used outside the class

app/presenters/scope_filters_presenter.rb
elenatanasoiu: :+1: used this

app/views/anonymous_feedback/explore/new.html.erb
elenatanasoiu: Dropped

spec/controllers/anonymous_feedback/document_types_controller_spec.rb
elenatanasoiu: Renamed to "makes a successful request"

spec/controllers/anonymous_feedback_controller_spec.rb
elenatanasoiu: Changed

spec/presenters/document_type_summary_presenter_spec.rb
elenatanasoiu: Hopefully corrected now. It might have been my RubyMine in the end, which was set to 4 space indentation when I started writing this spec. It is now set to 2 space.

spec/presenters/scope_filters_presenter_spec.rb
elenatanasoiu: Removed, it was supposed to test that the a string of comma separated paths would be parsed correctly, but we no longer expect a string. `clean_paths` will send us strictly arrays of values, never a string. I've removed this. It was an old test that got changed to pass without reading the description. 🤦‍♀️

spec/presenters/scope_filters_presenter_spec.rb
elenatanasoiu: 🤦‍♀️ Corrected

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -7,7 +7,7 @@ def index
     unless has_required_api_params?
       respond_to do |format|
         format.html { redirect_to anonymous_feedback_explore_url, status: 301 }
-        format.json { render json: { "errors" => ["Please set a valid 'path' or 'organisation' parameter"] }, status: 400 }
+        format.json { render json: { "errors" => ["Please provide a valid 'paths'/'path' or 'organisation' parameter"] }, status: 400 }
```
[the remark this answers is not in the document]
elenatanasoiu: Changed to comma

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -33,11 +34,20 @@ def export_request_params
...
+  end
+
+  def clean_paths
+    if params[:path]
+      params[:paths] = [params[:path]]
+    elsif params[:paths] && params[:paths].instance_of?(String)
+      params[:paths] = params[:paths].split(', ')
```
h-lame: The other clean_paths impl uses `split(',')` should we be consistent? FWIW I think the other one is "safer", we can always `split(',').map(&:strip)` if we're worried about leading or trailing spaces.

spec/controllers/anonymous_feedback/export_requests_controller_spec.rb
h-lame: :+1:

PR #368, opened 2018-01-15
Improvements to feedex

elenatanasoiu:
For: https://trello.com/c/amR8AiD5/309-feedex-changes-to-workflow

## Background 
While working on feedex, I requested an export for all the '/done' pages: https://support.integration.publishing.service.gov.uk/anonymous_feedback?path=%2Fdone

This export took 7 hours to perform and resulted in a 1.1 GB file. There was unfortunately no warning that the export would be so big, or that it would require a large amount of time. Had I received a warning, I would have thought twice about asking for such a large file. 

## What this does

Adds warning when attempting to export a search result.

### DIFF
--- a/app/helpers/feedex_helper.rb
+++ b/app/helpers/feedex_helper.rb
@@ -23,4 +23,12 @@ module FeedexHelper
     response_total = "Over #{response_total}" if results_limited
     response_total
   end
+
+  def confirmation_message(total_count, results_limited)
+    if results_limited
+      "This is a large number of records (more than #{total_count}) and may take some time.\n\nAre you sure you want to export?"
+    else
+      "This will export #{total_count} records.\n\nAre you sure you want to export?"
+    end
+  end
 end
--- a/app/views/anonymous_feedback/index.html.erb
+++ b/app/views/anonymous_feedback/index.html.erb
@@ -13,7 +13,7 @@
     <%= hidden_field_tag :to, @dates.to if @dates.to.present? %>
     <%= hidden_field_tag :path, @filtered_by.path if @filtered_by.path.present? %>
     <%= hidden_field_tag :organisation, @filtered_by.organisation_slug if @filtered_by.organisation_slug.present? %>
-    <%= submit_tag "Export as CSV", class: "btn-sm btn btn-default add-left-margin" %>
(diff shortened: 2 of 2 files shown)


app/helpers/feedex_helper.rb
```diff
@@ -23,4 +23,12 @@ def get_response_total(total_count, results_limited, scopes)
     response_total = "Over #{response_total}" if results_limited
     response_total
   end
+
+  def confirmation_message(total_count)
+    if total_count && total_count > 1000
+      "A lot of records are about to be exported: #{total_count}. Are you sure you to export this list?"
```
h-lame: Looks like we're missing a word: "Are you sure you _want_ to export this list?"

app/helpers/feedex_helper.rb
```diff
@@ -23,4 +23,12 @@ def get_response_total(total_count, results_limited, scopes)
...
+    else
+      "This will export #{total_count} records. Are you sure you want to export this list?"
+    end
```
h-lame: There's not much difference between these two messages, could we just have one message?  Or, always have "This will export <blah> records. Are you sure?" but add "This is a large number of records and may take some time." for the large case.

app/helpers/feedex_helper.rb
```diff
@@ -23,4 +23,12 @@ def get_response_total(total_count, results_limited, scopes)
     response_total = "Over #{response_total}" if results_limited
     response_total
   end
+
+  def confirmation_message(total_count)
+    if total_count && total_count > 1000
```
h-lame: Is 1000 records the limit where we want to start warning people that it's a lot of date?

There's something in feedex where it'll only paginate through `<some number>` of records and has a message saying "There are more records, use CSV export to see them all" on all the pages.  I wonder if whatever that number is and the number we use here should be the same number, or at least that the number we use here should be larger than that number.

elenatanasoiu: The max number of pages in `support-api` is set to 200, and the page size is 50, so the results are limited to 10.000. You can see the constants [here](https://github.com/alphagov/support-api/blob/master/app/models/anonymous_contact.rb#L54). I've changed the number to 10.000.

PR #382, opened 2018-02-01
Change campaign request form

elenatanasoiu:
For: https://trello.com/c/eOI3uArg/37-1-changes-to-support-form-for-campaign-requests

Adds new fields to the "Request a new campaign form" and reformulates questions. 

## Before 
before

## After

screenshot-support dev gov uk-2018 02 06-14-02-54

### DIFF
--- a/app/models/support/gds/campaign.rb
+++ b/app/models/support/gds/campaign.rb
@@ -4,14 +4,27 @@ module Support
   module GDS
     class Campaign
       include ActiveModel::Model
-      attr_accessor :title, :other_dept_or_agency, :signed_campaign, :start_date, :end_date, :description,
-                    :call_to_action, :success_measure, :proposed_url, :site_metadescription, :cost_of_campaign
+      attr_accessor :type_of_site, :has_read_guidance_confirmation, :has_read_oasis_guidance_confirmation,
+                    :signed_campaign, :start_date, :end_date, :development_start_date,
+                    :performance_review_contact_email, :government_theme, :description, :call_to_action,
+                    :proposed_url, :site_metadescription, :cost_of_campaign, :ga_contact_email
 
-      validates_presence_of :title, :signed_campaign, :start_date, :end_date, :description,
-                            :call_to_action, :success_measure, :proposed_url, :site_metadescription, :cost_of_campaign
-      validates_date :start_date, on_or_after: :today, on_or_before: :end_date
-      validates_date :end_date, on_or_after: :start_date
-      validates :proposed_url, format: /((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.campaign.gov.uk?/
+      validates_presence_of :type_of_site, :signed_campaign, :start_date, :end_date, :development_start_date,
+                            :performance_review_contact_email, :government_theme, :description, :call_to_action,
+                            :proposed_url, :site_metadescription, :cost_of_campaign, :ga_contact_email
+
+      validates_acceptance_of :has_read_guidance_confirmation, :has_read_oasis_guidance_confirmation, allow_nil: false
+      validates_inclusion_of :type_of_site, in: ['Campaign platform', 'Bespoke microsite']
+
+      validates_date :start_date, on_or_after: :today, before: :end_date
+      validates_date :end_date, after: :start_date
+      validates_date :development_start_date, on_or_after: :start_date
+      validates :proposed_url, format: /((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.(campaign\.)?gov.uk?/
+
+      validates_numericality_of :cost_of_campaign
+
+      VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
+      validates :performance_review_contact_email, :ga_contact_email, format: { with: VALID_EMAIL_REGEX }
     end
   end
 end
--- a/app/views/campaign_requests/_request_details.html.erb
+++ b/app/views/campaign_requests/_request_details.html.erb
@@ -2,29 +2,153 @@
   <p><%= f.object.class.description %></p>
 </div>
 
+<p>
+  Use this form for : Campaign Platform site or microsite
+<ul>
+  <li>A campaign platform site is a one page website on campaign.gov.uk</li>
+  <li>Microsite websites are not built on GOV.UK and may or may not have
+    a gov.uk url.
+    Microsite websites <%= link_to 'fall under Cabinet Office digital service expenditure', 'https://www.google.com/url?q=https://www.gov.uk/service-manual/agile-delivery/spend-controls-check-if-you-need-approval-to-spend-money-on-a-service%23when-your-service-is-considered-a-digital-service&sa=D&ust=1518444538536000&usg=AFQjCNFvnZ81hn7cqJnGHgcgQ0V0bdwCBQ'%>
+    therefore GDS spend approval must be sought following approval from GCS to develop a campaign microsite.
+  </li>
+</ul>
+  For either type of site, please answer the questions below.
+</p>
+
 <%= f.semantic_fields_for :campaign do |r| %>
   <%= r.inputs :name => "Campaign details" do %>
-    <%= r.input :title, label: "Campaign title", required: true, input_html: {:class => "input-md-6", :"aria-required" => true} %>
+    <%= r.input :type_of_site,
+                label: "Are you applying for the campaign platform or a bespoke microsite?",
+                as: :radio,
+                :collection => ["Campaign platform", "Bespoke microsite"],
+                required: true,
+                input_html: {:class => "input-md-6"} %>
+    <%= r.input :has_read_guidance_confirmation,
+                label: "<b>Have you read the the #{link_to('GCS guidance on campaign websites', 'https://gcs.civilservice.gov.uk/guidance/campaigns/websites-government-campaigns/')} and accept the requirements for a Campaign Platform website?</b>".html_safe,
+                as: :boolean,
+                input_html: {:class => "input-md-6"} %>
+    <%= r.input :has_read_oasis_guidance_confirmation,
+                label: "<b>Have you followed the #{link_to('GCS guidance for OASIS planning', 'https://gcs.civilservice.gov.uk/wp-content/uploads/2015/09/6.3938_CO_GCS-Campaign-Planning_FINAL_A4_111017.pdf')} and are using the #{link_to('GCS website recommended OASIS template?', 'https://gcs.civilservice.gov.uk/guidance/campaigns/strategy-planning-templates/')}</b>".html_safe,
+                as: :boolean,
+                input_html: {:class => "input-md-6"} %>
+    <%= r.input :signed_campaign,
+                label: "Name of the head of digital who signed off the campaign website application",
+                required: true,
+                input_html: {:class => "input-md-6"} %>
+    <%= r.input :start_date,
+                label: "Start date of campaign site",
+                required: true,
+                input_html: { class: "input-md-2",
+                              :"data-module" => 'calendar',
+                              :"data-format" => 'dd-mm-yy',
+                              :"data-min-date" => 0,
+                              :placeholder => "dd-mm-yyyy",
+                              :value => r.object.start_date} %>
+    <%= r.input :end_date,
+                label: "Proposed end date of campaign site",
+                required: true,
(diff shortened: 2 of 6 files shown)


app/models/support/gds/campaign.rb
```diff
@@ -4,14 +4,22 @@ module Support
...
-      attr_accessor :title, :other_dept_or_agency, :signed_campaign, :start_date, :end_date, :description,
-                    :call_to_action, :success_measure, :proposed_url, :site_metadescription, :cost_of_campaign
+      attr_accessor :type_of_site, :has_read_guidance, :has_read_oasis_guidance, :signed_campaign,
+                    :start_date, :end_date, :development_start_date, :performance_review_contact_email,
+                    :government_theme, :description, :call_to_action, :proposed_url, :site_metadescription,
+                    :cost_of_campaign, :ga_contact_email
 
-      validates_presence_of :title, :signed_campaign, :start_date, :end_date, :description,
-                            :call_to_action, :success_measure, :proposed_url, :site_metadescription, :cost_of_campaign
+      validates_presence_of :type_of_site, :has_read_guidance, :has_read_oasis_guidance, :signed_campaign,
```
h-lame: Given we present a radio box with only two choices to the user could we validate that `type_of_site` only allows those two values?

h-lame: Do we need to store the `has_read_guidance` and `has_read_oasis_guidance` checkboxes?  Are they intended as gates were saying no means you can't submit the form (e.g. like checking "I have read the terms & conditions", or "I have read the EULA" we might get on signing up to a new website, or installing new software)?  We could use rails validates_confirmation_of on those attributes to not proceed if they don't tick yes.  (The doc isn't clear on this, so maybe it's ok to let it through if they haven't read those docs).

app/views/campaign_requests/_request_details.html.erb
```diff
@@ -2,26 +2,148 @@
...
+                             :"aria-describedby" => "seoHelpBlock" } %>
+    <p class="help-block">
+      The meta description is a short, clear sentence how you want the site described as it appears in Google search
+    </p>
 
-    <%= r.input :site_metadescription, as: :text, label: "Site metadescription (appears in search results)", required: true, input_html: {:class => "input-md-6", :rows => 6, :cols => 50, :"aria-required" => true, :"aria-describedby" => "seoHelpBlock" } %>
+    <%= r.input :cost_of_campaign, label: "Site build budget / costs (and overall campaign cost, if applicable)", required: true, input_html: { class: "input-md-6" } %>
+    <p class="help-block">
+      If you are applying for the Campaign Platform page, this is free of any build charge
+    </p>
```
h-lame: As this bit of the form is only relevant to people wanting a bespoke site, are we able to hide it for campaign sites and only show it to those choosing the bespoke site option?  It'd mean we wouldn't need this help block to explain to people choosing a campaign site that they can leave this blank?  But maybe that's intentional, as a pointer to make people choose campaign instead of bespoke?

app/views/campaign_requests/_request_details.html.erb
```diff
@@ -2,26 +2,148 @@
...
+                             :"aria-required" => true } %>
 
-    <%= r.input :call_to_action, as: :text, label: "Call to action", required: true, input_html: {:class => "input-md-6", :rows => 6, :cols => 50, :"aria-required" => true } %>
-    <%= r.input :success_measure, as: :text, label: "How will you measure success?", required: true, input_html: {:class => "input-md-6", :rows => 6, :cols => 50, :"aria-required" => true } %>
+    <%= r.input :proposed_url,
+                label: "Proposed URL (in the form of xxxxx.campaign.gov.uk)",
+                required: true, input_html: { class: "input-md-6",
+                                              :"aria-describedby" => "urlHelpBlock",
+                                              :"aria-required" => true } %>
+    <p class="help-block">
```
h-lame: Do we need a `proposed_url_reasons` field for when they've asked for something that's not `xxxx.campaign.gov.uk`, or `xxxx.gov.uk` ?  The help text suggests stating their reasons, but they can't put those reasons in the `proposed_url` field as then it won't validate as a url.

app/models/support/gds/campaign.rb
```diff
@@ -4,14 +4,22 @@ module Support
...
+                            :start_date, :end_date, :development_start_date, :performance_review_contact_email,
+                            :government_theme, :description, :call_to_action, :proposed_url, :site_metadescription,
+                            :cost_of_campaign, :ga_contact_email
       validates_date :start_date, on_or_after: :today, on_or_before: :end_date
       validates_date :end_date, on_or_after: :start_date
+      validates_date :development_start_date, on_or_after: :start_date
       validates :proposed_url, format: /((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.campaign.gov.uk?/
```
h-lame: Might need to change this regex because the form now suggests `xxxx.campaign.gov.uk` for campaigns (covered by this regex) and  `xxxx.gov.uk` for bespoke (not covered by this regex).  Perhaps we could use a conditional validator based on the value of `type_of_site`.  But we also state that you can choose something else if you give us reasons so we need some other way of validating it.

app/models/support/gds/campaign.rb
elenatanasoiu: Those two fields correspond to checkboxes in the form, where I'm using `required: true` to ask people to tick them. Once ticked, we're stripping them out in the CampaignRequestsController so they don't need to be forwarded. 

So we don't require the services of `validates_confirmation_of` since we never assign those fields. 

However, I've renamed the fields to `has_read_guidance_confirmation` and `has_read_oasis_guidance_confirmation` to make it more clear what they are.

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Changed to make the `campaign.` part optional: 
`/((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.(campaign\.)?gov.uk?/`

spec/models/support/gds/campaign_spec.rb
```diff
@@ -8,38 +8,69 @@ def as_str(date)
...
       it { should validate_presence_of(:site_metadescription) }
       it { should validate_presence_of(:cost_of_campaign) }
+      it { should validate_presence_of(:ga_contact_email) }
 
-      it { should allow_value("example.campaign.gov.uk").for(:proposed_url) }
-      it { should_not allow_value("mycampaign.gov.uk").for(:proposed_url) }
 
-      it { should_not allow_value("xxx").for(:start_date) }
-      it { should_not allow_value("xxx").for(:end_date) }
+      it { should not(allow_value('Non Campaign platform').for(:proposed_url)) }
```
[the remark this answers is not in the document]
elenatanasoiu: Changed to `should_not`

app/models/support/gds/campaign.rb
elenatanasoiu: Changed to make the `campaign.` part optional: 
`/((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.(campaign\.)?gov.uk?/`

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Will ask.

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Will ask.

app/models/support/gds/campaign.rb
h-lame: Arg.  I meant `validates_acceptance_of`, not `validates_confirmation_of`.  I don't *think* we can get away with `required: true` because rails checkboxes always submit a `0` or `1` so they're always filled in.  There's a hidden input that always submits the `0`, and if you check it the form also submits the `1` overwriting the `0` earlier in the form.  Of course, this assumes we're using rails checkbox helpers in the form which behave this way, maybe we're not.

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Have asked. We're keeping it for both types of websites as it's a way for them to keep track of costs.

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Have asked. We've removed the part of the help text that suggests you have any other option besides `xxxxx.campaign.gov.uk` or `xxxxx.gov.uk`.

app/models/support/gds/campaign.rb
```diff
@@ -4,14 +4,29 @@ module Support
...
-      validates_date :end_date, on_or_after: :start_date
-      validates :proposed_url, format: /((http|https):\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.campaign.gov.uk?/
+      validates_presence_of :type_of_site, :signed_campaign, :start_date, :end_date, :development_start_date,
+                            :performance_review_contact_email, :government_theme, :description, :call_to_action,
+                            :proposed_url, :site_metadescription, :cost_of_campaign, :ga_contact_email
+
+      validates_acceptance_of :has_read_guidance_confirmation, :has_read_oasis_guidance_confirmation, allow_nil: false
+      validates_inclusion_of :type_of_site, in: ['Campaign platform', 'Bespoke microsite']
+
+      validates_acceptance_of :has_read_guidance_confirmation, :has_read_oasis_guidance_confirmation, allow_nil: false
```
h-lame: This line is here twice, I don't think that means it'll have two errors when you forget to check the button, but it might.  Even if rails does the right thing, we should delete one version to avoid confusion when we change one line but not the other.

app/views/campaign_requests/_request_details.html.erb
```diff
@@ -2,29 +2,153 @@
...
 
-    <%= r.input :cost_of_campaign, label: "Cost of campaign", required: true, input_html: { class: "input-md-6" } %>
+    <%= r.input :ga_contact_email, label: "Contact details for Google Analytics leads (Gmail accounts only)", required: true, input_html: {:class => "input-md-6", :rows => 6, :cols => 50, :"aria-required" => true } %>
+    <p class="help-block">
+      GCS centralise and coordinate access to Google Analytics, although departments will themselves
+      be responsible for analysing their site’s performance.
+    </p>
+    <p class="help-block">
+      If you need to set up a bespoke analytics package independent of centralised/GCS GA for the purposes
+      of your marketing, please ​provide the following with access ​webmaster@cabinetoffice.gov.uk &
```
h-lame: Github has flagged this `&` in red.  Perhaps we should change it to either `and` or `&amp;`.

app/views/campaign_requests/_request_details.html.erb
```diff
@@ -2,29 +2,153 @@
...
+      microsite​s​@cabinetoffice.gov.uk.
+    </p>
 
     <% end %>
 <% end %>
 
 <%= f.input :additional_comments, as: :text, label: "Additional comments", input_html: { class: "input-md-6", rows: 6, cols: 50 } %>
 
+<p class="help-block">
+  Please attach your OASIS plan using this template (<%= link_to 'OASIS template from the GCS website', 'https://gcs.civilservice.gov.uk/guidance/campaigns/strategy-planning-templates/' %>). In particular you need to demonstrate how the digital / website element of the campaign integrates into your overall OASIS plan and we will need to see <u>specific digital KPIs</u> in the evaluation section.
```
h-lame: The wording here seems likely to confuse users.  "Please attach your OASIS plan" suggests to me that there will be a file upload function, but there isn't.  It seems there's some next stage after this form where the user can attach their OASIS template so we should make it clear that they should have their OASIS plan ready to attach later, but not make them think they can attach it to this form.

elenatanasoiu: I've changed the text to "Please note that in the next stage you will also need to provide an OASIS plan using this template" and gotten approval for it.

app/views/campaign_requests/_request_details.html.erb
elenatanasoiu: Changed to "and"

app/models/support/gds/campaign.rb
elenatanasoiu: Removed one

app/views/campaign_requests/_request_details.html.erb
h-lame: :+1:

PR #498, opened 2018-09-21
[DO NOT MERGE] Add Content Publisher to technical fault report

emmabeynon:
For https://trello.com/c/sDIsMglD/258-add-content-publisher-to-list-of-apps-in-support-form

This PR adds Content Publisher as an option to select when raising a technical support request:

screen shot 2018-09-21 at 17 23 42

We also add a `assignee_id` attribute to the payload sent to the Zendesk API when creating a document (see Zendesk API documentation to [create a ticket](https://developer.zendesk.com/rest_api/docs/core/tickets#create-ticket) and [accepted attributes](https://developer.zendesk.com/rest_api/docs/core/tickets#json-format)).  When Content Publisher is in private beta we'd like for any support tickets raised relating to it to be automatically assigned to the Product Manager responsible for it, which is why we've hardcoded their Zendesk ID.  When the app is no longer in private beta we expect this to be reverted.

There isn't a staging environment available for Zendesk so we will only know if this approach to assigning an agent works correctly once it's been deployed to production.  In the meantime we have tested that the payload is accepted by the [gds_zendesk dummy client](https://github.com/alphagov/gds_zendesk/commit/14c28f97ffd27dfaba6c4f466ca7272e29f48fc6).

### DIFF
--- a/app/controllers/technical_fault_reports_controller.rb
+++ b/app/controllers/technical_fault_reports_controller.rb
@@ -16,7 +16,7 @@ protected
   def technical_fault_report_params
     params.require(:support_requests_technical_fault_report).permit(
       :fault_context, :fault_specifics, :actions_leading_to_problem,
-      :what_happened, :what_should_have_happened,
+      :what_happened, :what_should_have_happened, :assignee_id,
       requester_attributes: %i[email name collaborator_emails],
       fault_context_attributes: %i[name id inside_government_related],
     ).to_h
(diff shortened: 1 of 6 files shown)


benthorner:
@emmabeynon would it make sense to close this for the time being?

emmabeynon:
@benthorner I've agreed with Humin to keep this open and merge when the app is available to publishers.  He needs to do some Zendesk admin before this gets merged.

benthorner:
Superseded by https://github.com/alphagov/support/pull/524.

PR #521, opened 2018-11-16
Putting in a new support feedback page for the content publisher

kgarwood:
Putting in a new support feedback page for the content publisher.

These show the changes that were made to the support app to add a new support page.  Wherever possible in the code and the tests, I've borrowed conventions from how the other pages.

### DIFF
--- a/app/controllers/content_publisher_feedback_requests_controller.rb
+++ b/app/controllers/content_publisher_feedback_requests_controller.rb
@@ -1,0 +1,24 @@
+class ContentPublisherFeedbackRequestsController < RequestsController
+protected
+
+  def new_request
+    Support::Requests::ContentPublisherFeedbackRequest.new
+  end
+
+  def zendesk_ticket_class
+    Zendesk::Ticket::ContentPublisherFeedbackRequestTicket
+  end
+
+  def parse_request_from_params
+    Support::Requests::ContentPublisherFeedbackRequest.new(content_publisher_feedback_request_params)
+  end
+
+  def content_publisher_feedback_request_params
+    params.require(:support_requests_content_publisher_feedback_request).permit(
+      :feedback_type,
+      :feedback_details,
+      :impact_on_work,
+      requester_attributes: %i[email name collaborator_emails],
+    ).to_h
+  end
+end
--- a/app/models/support/requests/content_publisher_feedback_request.rb
+++ b/app/models/support/requests/content_publisher_feedback_request.rb
@@ -1,0 +1,31 @@
+require 'active_support/core_ext'
+
+module Support
+  module Requests
+    class ContentPublisherFeedbackRequest < Request
+      attr_accessor :feedback_type, :feedback_details, :impact_on_work
+
+      OPTIONS = {
+        "accessibility or usability" => "accessibility or usability",
+        "working unexpectedly" => "something working differently than you expected",
+        "helpful feature" => "a feature that would help you with your work",
+        "other" => "something else",
+      }.freeze
+
+      validates_presence_of :feedback_type, :feedback_details
+      validates :feedback_type, inclusion: { in: OPTIONS.keys }
+
+      def self.label
+        "Give feedback on Content Publisher (Beta)"
+      end
+
+      def self.description
+        "Suggest changes to features within Content Publisher."
+      end
+
+      def feedback_type_options
+        OPTIONS.map { |key, value| [value, key] }
+      end
+    end
+  end
+end
--- a/app/views/content_publisher_feedback_requests/_request_details.html.erb
+++ b/app/views/content_publisher_feedback_requests/_request_details.html.erb
@@ -1,0 +1,15 @@
+<div class="alert alert-info">
+  <p>Content Publisher is in private beta and supports most of the features that are available on Whitehall.</p>
+  <p>With this form you can tell us more about your experience with the new app so we can make it better. If you’re having issues publishing or editing a document raise a <a class="link-inherit" href="<%= new_technical_fault_report_path %>">support request</a> so you can get technical help.</p>
+</div>
+<div id="feedback-type">
+  <%= f.input :feedback_type, as: :radio, required: true, label: "Your feedback is about", collection: f.object.feedback_type_options, input_html: { :"aria-required" => true } %>
+  <%= f.input :feedback_details, as: :text, required: true,
+              label: "Tell us a bit more",
+              input_html: { class: "input-md-6", rows: 6, cols: 50, :"aria-required" => true } %>
+  <%= f.input :impact_on_work, as: :text, required: false,
+              label: "What's the impact on your work if we don't do anything about it?",
+              hint: "This information will help us prioritise changes and future work",
+              input_html: { class: "input-md-6", rows: 6, cols: 50, :"aria-required" => false } %>
+</div>
+<%= render partial: "support/collaborators", locals: { f: f } %>
--- a/spec/models/support/requests/content_publisher_feedback_request_spec.rb
+++ b/spec/models/support/requests/content_publisher_feedback_request_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+module Support
+  module Requests
+    describe ContentPublisherFeedbackRequest do
+      def request(options = {})
+        ContentPublisherFeedbackRequest.new(options).tap(&:valid?)
+      end
+
+      it { should validate_presence_of(:feedback_type) }
+      it { should validate_presence_of(:feedback_details) }
+
+      it { should allow_value("accessibility or usability").for(:feedback_type) }
+      it { should allow_value("working unexpectedly").for(:feedback_type) }
+      it { should allow_value("helpful feature").for(:feedback_type) }
+      it { should allow_value("other").for(:feedback_type) }
+      it { should_not allow_value("xxx").for(:feedback_type) }
+
+      it { should allow_value("").for(:impact_on_work) }
+
+      it "provides type of change choices" do
+        expect(request.feedback_type_options).to_not be_empty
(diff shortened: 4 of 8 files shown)


app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,32 @@
...
+      def self.description
+        "Suggest changes to features within the content publisher."
+      end
+
+      def feedback_type_options
+        [
+          ["an accessibility issue or a usability issue", "accessibility_or_usability"],
+          ["something working differently than you expected", "working_unexpectedly"],
+          ["a feature that would help you with your work", "helpful_feature"],
+          ["something else", "other"],
```
kevindew: Do we need these values to be snake case? - It looks like it produces a somewhat hard to read output in the email: https://github.com/alphagov/support/pull/521/files#diff-4f144761556ba2db8112558bccb3d2ecR19 and I assume we just need these to be strings?

app/views/content_publisher_feedback_requests/_request_details.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="alert alert-info">
+  <p>Content Publisher is in private beta and supports most of the features that are available on Whitehall.</p>
```
kevindew: Note that I see here we describe "Content Publisher" with title case, whereas the other bits of content in this PR seems to have it as sentence case: "Content publisher" - do we know which way we're supposed to use?

app/views/content_publisher_feedback_requests/_request_details.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p>With this form you can tell us more about your experience with the new app so we can make it better. If you’re having issues publishing or editing a document raise a <a class="govuk-link govuk-link--no-visited-state" href="<%= Plek.new.external_url_for("support") + "/technical_fault_report/new" %>">support request</a> so you can get technical help.</p>
```
kevindew: As this link is displayed within support we don't need to use plek here as you are already on that hostname.

We should also link to this via route rather than hard coding the URL.

Also "govuk-link govuk-link--no-visited-state" has no effect in this application as [govuk-frontend](https://github.com/alphagov/govuk-frontend) is not part of this app.

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,32 @@
+require 'active_support/core_ext'
+
+module Support
+  module Requests
```
kevindew: We seem to be missing the unit tests for this class

app/controllers/content_publisher_feedback_requests_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  def parse_request_from_params
+    Support::Requests::ContentPublisherFeedbackRequest.new(content_publisher_feedback_request_params)
+  end
+
+  def content_publisher_feedback_request_params
+    params.require(:support_requests_content_publisher_feedback_request).permit(
+      :feedback_type,
+      :feedback_details,
+      :impact_on_work,
+      requester_attributes: %i[email name collaborator_emails],
```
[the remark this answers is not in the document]
kgarwood: Done.  The partial is included to make it behave like more of the other support forms.

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,32 @@
...
+    class ContentPublisherFeedbackRequest < Request
+      attr_accessor :feedback_type, :feedback_details, :impact_on_work
+
+      validates_presence_of :feedback_type
+
+      def initialize(attrs = {})
+        super(attrs)
+      end
```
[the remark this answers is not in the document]
kgarwood: Done.

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,32 @@
+require 'active_support/core_ext'
+
+module Support
+  module Requests
+    class ContentPublisherFeedbackRequest < Request
+      attr_accessor :feedback_type, :feedback_details, :impact_on_work
+
+      validates_presence_of :feedback_type
```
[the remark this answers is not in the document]
kgarwood: We now check for the presence of feedback_details. impact_on_work remains optional.  The required/optional aspect of these fields is also reflected in the view for the form.

app/models/support/requests/content_publisher_feedback_request.rb
kgarwood: Done. Snake case removed.  I'll make a note more broadly on the repo to suggest we don't need this in the other forms.

app/models/support/requests/content_publisher_feedback_request.rb
kgarwood: Done. We now have unit tests for this request.

app/views/content_publisher_feedback_requests/_request_details.html.erb
kgarwood: Done. govuk-link govuk-link--no-visited-state has been removed.  The href part now uses a rails path: href="<%= new_technical_fault_report_path %>"> and it works.

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,32 @@
...
+      def initialize(attrs = {})
+        super(attrs)
+      end
+
+      def self.label
+        "Provide feedback on content publisher"
+      end
+
+      def self.description
+        "Suggest changes to features within the content publisher."
```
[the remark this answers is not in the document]
kgarwood: Done

app/views/content_publisher_feedback_requests/_request_details.html.erb
kgarwood: All references in the repo now use "Content Publisher".  Checked this with content developers.

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,30 @@
+require 'active_support/core_ext'
+
+module Support
+  module Requests
+    class ContentPublisherFeedbackRequest < Request
+      attr_accessor :feedback_type, :feedback_details, :impact_on_work
+
+      validates_presence_of :feedback_type, :feedback_details
+      validates :feedback_type, inclusion: { in: ["accessibility or usability", "working unexpectedly", "helpful feature", "other"] }
```
kevindew: Rather than repeating these can we have a single definition to check them? - Maybe have feedback_type_options as a constant?

app/views/content_publisher_feedback_requests/_request_details.html.erb
```diff
@@ -0,0 +1,15 @@
...
+</div>
+<div id="feedback-type">
+  <%= f.input :feedback_type, as: :radio, required: true, label: "Your feedback is about", collection: f.object.feedback_type_options, input_html: { :"aria-required" => true } %>
+  <%= f.input :feedback_details, as: :text, required: true,
+              label: "Tell us a bit more",
+              input_html: { class: "input-md-6", rows: 6, cols: 50, :"aria-required" => true } %>
+  <%= f.input :impact_on_work, as: :text, required: false,
+              label: "What's the impact on your work if we don't do anything about it?",
+              hint: "This information will help us prioritise changes and future work",
+              input_html: { class: "input-md-6", rows: 6, cols: 50, :"aria-required" => true } %>
```
kevindew: This should be aria-required false right if we're not requiring this field?

spec/models/support/requests/content_publisher_feedback_request_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      it { should allow_value("other").for(:feedback_type) }
+      it { should_not allow_value("xxx").for(:feedback_type) }
+
+      it { should allow_value("").for(:impact_on_work) }
+
+      it "provides type of change choices" do
+        expect(request.feedback_type_options).to_not be_empty
+      end
+    end
+  end
```
kevindew: This looks good 👍

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,41 @@
...
+        OPTIONS.map { |key, value| [value, key] }
+      end
+      # def feedback_type_options
+      #   [
+      #     ["accessibility or usability", "accessibility or usability"],
+      #     ["something working differently than you expected", "working unexpectedly"],
+      #     ["a feature that would help you with your work", "helpful feature"],
+      #     ["something else", "other"],
+      #   ]
+      # end
```
kevindew: You'll want to remove this

app/models/support/requests/content_publisher_feedback_request.rb
```diff
@@ -0,0 +1,41 @@
...
+        "working unexpectedly" => "something working differently than you expected",
+        "helpful feature" => "a feature that would help you with your work",
+        "other" => "something else",
+      }.freeze
+
+      validates_presence_of :feedback_type, :feedback_details
+      validates :feedback_type, inclusion: { in: OPTIONS.keys }
+
+
+
```
kevindew: Looks like we have some superfluous new lines to remove

app/models/support/requests/content_publisher_feedback_request.rb
kgarwood: Done

app/models/support/requests/content_publisher_feedback_request.rb
kgarwood: Done. Spaces removed, it's interesting to note how line spaces are done from request to request.

PR #573, opened 2019-02-26
Add linting to the default rake task

peteglondon:
The lint task is not run by default.

This means you don't find out about linting errors until CI.

We want to `fail fast` so adding linting as part of the default
rake task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,4 +6,10 @@ require File.expand_path('../config/application', __FILE__)
 
 Support::Application.load_tasks
 
-task :default => [:spec, 'jasmine:ci']
+Rake::Task['default'].clear
+
+task :default do
+  Rake::Task['lint'].invoke
+  Rake::Task['spec'].invoke
+  Rake::Task['jasmine:ci'].invoke
+end


lib/tasks/lint.rake
```diff
@@ -1,3 +1,5 @@
+task default: :lint
```
cbaines: How will this work with the configuration in the Rakefile?

https://github.com/alphagov/support/blob/master/Rakefile#L9

peteglondon: These will still run, I had to do it this way here https://github.com/alphagov/finder-frontend/pull/814. I think because it didn't find the lint task when I added it to the Rakefile.

Rakefile
```diff
@@ -6,4 +6,4 @@ require File.expand_path('../config/application', __FILE__)
 
 Support::Application.load_tasks
 
-task :default => [:spec, 'jasmine:ci']
+task :default => [:spec, 'jasmine:ci', :lint]
```
DilwoarH: Does it run in this order? if so, would it make sense for it to do lint first so it fails before running tests?

peteglondon: It doesn't seem to run in this order.

peteglondon: Managed to make it run in the order we want by clearing the default rake task.

DilwoarH: Great 👍

DilwoarH:
LGTM 👍

PR #588, opened 2019-03-28
[Do not merge] Update content support form to match user-facing facet values

leenagupte:
Trello: https://trello.com/c/xzww3I5e

## What's changed
In order to add a page to the business finder, publishers fill in a form which creates a Zendesk ticket that goes to content support.

The fields that appear in the Zendesk ticket show 'values' which do not match the facet values that are in the new content tagger interface for the business finder.

The fields that appear in Zendesk should be the same as the fields that appear in the content support form which should also be the same as the fields that are in the content tagger interface.

### Before
https://govuk.zendesk.com/agent/tickets/3596718
```
[Type] 
adding content

[Url] 
/guidance/some-guidance

[Sector]

gas-markets

[Business activity]

[Employing eu citizens] 
no

[Personal data] 
no

[Intellectual property]

[Funding schemes]

[Public sector procurement]
```

### After
https://govuk.zendesk.com/agent/tickets/3642774
```
[Type]
Adding content to the finder

[Url]
/guidance/some-guidance

[Pinned content]
Yes

[Explanation]
Testing changes to how the Zendesk ticket is created.

[Sector]

Oil, gas and coal

[Business activity]

[Employing eu citizens]
No

[Personal data]
No

[Intellectual property]

[Funding schemes]

[Public sector procurement]
```

## Next Steps
Rather than hardcoding the list of sectors and business activity, etc, they should be read in from publishing-api.


app/views/eu_exit_business_readiness_requests/_request_details.html.erb
```diff
@@ -66,15 +66,15 @@
   :sector,
   as: :check_boxes,
   label: "What sector(s) is the content is about?",
-  collection: f.object.sector_options,
+  collection: Support::Requests::EuExitBusinessReadinessRequest::SECTOR_OPTIONS,
```
vanitabarrett: This should be part of the first commit as it's related to sectors, not business activity.

leenagupte: Fixed and fixed up. 😃

vanitabarrett: LGTM! 👍

PR #591, opened 2019-03-29
Get brexit feedback

JonathanHallam:
As it stands there's a bug in the support app which makes it difficult to get feedback for a large number of endpoints at the same time that's quite a lot of work to fix. This is a relatively quick solution that pulls endpoints from google spreadsheets and compares the endpoints therein to feedback from the support api. I fully expect this to be deleted once that bug is fixed.

This will also require a couple of things to be put into the govuk-secrets repo which I don't have access to.


vanitabarrett:
Is there a reason why the list of endpoints needs to be stored in Google sheets? Could it instead be, for example, a YAML file that we add here and read in?

JonathanHallam:
@vanitabarrett It's in sheets at the moment because it's a mixture of endpoints that are collected programmatically and updated manually by non devs. Major issue is that it's a list that changes semi regularly and is edited by people that don't use git.

spec/controllers/brexit_feedback_controller_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+describe BrexitFeedbackController, type: :controller do
```
steventux: Each controller action should be tested for response types here.

spec/models/support/requests/brexit_feedback_request_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+require "gds_api/support_api"
+require 'gds_api/test_helpers/support_api'
+
+module Support
+  module Requests
+    include GdsApi::TestHelpers::SupportApi
+
+    describe BrexitFeedbackRequest do
```
steventux: This tests the happy path, given we use a lot of external services in this PR we should also test what happens with common endpoint failure cases.

app/assets/response_1.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
JonathanHallam: Heya, thanks for the review, I'll work through it now. These files are used to test how the responses from the support-api are paginated. Is there a way that mocks are usually handled? The integration addresses are just there because that's where the original data comes from but they're not used in the tests.

app/controllers/brexit_feedback_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+class BrexitFeedbackController < ApplicationController
+  def home
+    session.destroy
+  end
+
+  def results
+    client_hash = session['auth_client']
+
+    opts = {
```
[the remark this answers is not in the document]
JonathanHallam: I think some of this could probably be memoised but as it stands the use case for this is quite small. Essentially it's probably only going to be run once a week before being edited manually so I've just done this for clarity.

app/models/support/requests/brexit_feedback_request.rb
```diff
@@ -0,0 +1,78 @@
...
+        @brexit_slugs = slugs
+        @support_api = support_api
+        @first_page = get_first_page
+        @formatted_results = format_results(@first_page).reject do |res|
+          is_brexity?(res["path"]) == false
+        end
+        get_subsequent_pages
+      end
+
+      def support_api
```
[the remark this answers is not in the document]
JonathanHallam: 👍

app/assets/response_1.json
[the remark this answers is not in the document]
steventux: Usually test fixtures like this go in the `spec/fixtures` or `test/fixtures` directory. You may need to add something in your tests to load them.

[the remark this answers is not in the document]
JonathanHallam: cool thanks, will move them over

app/controllers/brexit_feedback_controller.rb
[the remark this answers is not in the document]
steventux: The usual goal with Rails controller code is to keep as small a code footprint as possible.
Calling a service class or PORO to create this hash would be preferable.
It's also easier to unit test.

app/models/support/requests/brexit_feedback_request.rb
```diff
@@ -0,0 +1,78 @@
...
+      def get_subsequent_pages
+        page = 2
+        until page >= @first_page["pages"] + 1 do
+          results = @support_api.problem_reports(from_date: @from_date, to_date: @to_date, page: page).to_hash
+          @formatted_results += format_results(results)
+          page += 1
+        end
+      end
+
+      def format_results(result_hash)
```
steventux: I don't think this belongs here. It's a response formatter and would be better split out into a separate class.

app/controllers/brexit_feedback_controller.rb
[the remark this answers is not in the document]
JonathanHallam: okay, thanks, I'll clean this up some

app/models/support/requests/brexit_slug_fetcher.rb
```diff
@@ -0,0 +1,40 @@
...
+        response = Net::HTTP.get_response(uri)
+        JSON.parse(response.body)["results"].map { |page| page["link"] }
+      end
+
+      def get_google_sheets_slugs
+        service = Google::Apis::SheetsV4::SheetsService.new
+        service.client_options.application_name = APPLICATION_NAME
+        service.authorization = @auth_client
+
+        spreadsheet_id = "1bFSDYFT5fBpDQTvAeqw4j7QhYXTnFmDuGCLGDwx-wYk"
```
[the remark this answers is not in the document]
JonathanHallam: 👍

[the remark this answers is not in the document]
JonathanHallam: 👍

client_secrets.json
```diff
@@ -0,0 +1,12 @@
+{
+  "web": {
+    "client_id": "client_id",
+    "project_id": "managing-feedback",
+    "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+    "token_uri": "https://oauth2.googleapis.com/token",
+    "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+    "client_secret": "client_secret",
+    "redirect_uris": ["http://support.dev.gov.uk/brexit/results", "https://signon.staging.publishing.service.gov.uk/brexit/results", "https://signon.integration.publishing.service.gov.uk/brexit/results", "https://signon.publishing.service.gov.uk/brexit/results"],
```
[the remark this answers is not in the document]
JonathanHallam: 👍

app/models/support/requests/brexit_slug_fetcher.rb
```diff
@@ -0,0 +1,40 @@
...
+        range = 'List of documents with facets!A2:A'
+        response = service.get_spreadsheet_values(spreadsheet_id, range)
+
+        spreadsheet_id = "1Pzjeae33PQKzekCJfx95VYlNNPH6ePhsQSMFEIrsY5g"
+        range = 'Brexit docs!A4:A'
```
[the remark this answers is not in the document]
JonathanHallam: 👍

app/models/support/requests/brexit_slug_fetcher.rb
```diff
@@ -0,0 +1,40 @@
...
+      attr_accessor :slugs, :auth_client
+      def initialize(auth_client)
+        @auth_client = auth_client
+        @slugs = get_taxon_slugs + get_google_sheets_slugs
+      end
+
+    private
+
+      def get_taxon_slugs
+        uri = URI("https://www.gov.uk/api/search.json?filter_taxons=d7bdaee2-8ea5-460e-b00d-6e9382eb6b61&count=50&fields=description&fields=link&fields=title")
```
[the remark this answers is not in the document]
JonathanHallam: 👍

client_secrets.json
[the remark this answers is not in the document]
JonathanHallam: I don't think so, as I understand this file will be pretty static and I don't think this function will be especially long lived anyway.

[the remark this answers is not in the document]
steventux: Sorry, I wasn't very clear the first time round. Should this file be committed to version control? It serves no purpose in a production environment.
Will it be overwritten in the deployment process?
I would suggest `config/` as an appropriate location for this data in any case.

app/controllers/brexit_feedback_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    auth_client = Signet::OAuth2::Client.new(opts)
+
+    auth_client.code = params['code']
+    auth_client.additional_parameters = { access_type: 'offline', include_granted_scopes: 'true' }
+    brexit_slugs = Support::Requests::BrexitSlugFetcher.new(auth_client).slugs
+    @results = Support::Requests::BrexitFeedbackRequest.new(Date.parse(session['from_date']), Date.parse(session['to_date']), brexit_slugs).formatted_results
+  end
+
+  def auth
+    secrets = JSON.parse(File.read('client_secrets.json'))
```
steventux: It might be simpler to make this file an JSON ERB template.
Then you can interpolate environment variables as you read the file and there's no need to rewrite the file:
`config/client_secrets.json.erb`:

```
{
  "web": {
    "client_id": <%= ENV.fetch("google_client_id", "client_id") %>,
    "client_secret": <%= ENV.fetch("google_client_secret", "client_secret") %>,
...
}
```

And then you can read/interpolate vars like so:

```ruby
template = File.read('config/client_secrets.json')
secrets = JSON.parse(ERB.new(template).results)
```

app/controllers/brexit_feedback_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    secrets["web"]["client_id"] = ENV["google_client_id"]
+    secrets["web"]["client_secret"] = ENV["google_client_secret"]
+    File.open("client_secrets.json", "w+") do |f|
```
steventux: Does this `#auth` method get called for every request? It's doing IO repeatedly on the same file, I wonder if there's a more efficient way of handling this. Can the authorisation be stored in memory once generated?

app/controllers/brexit_feedback_controller.rb
JonathanHallam: nice, didn't know this was a thing. Thanks :)

JonathanHallam: So the google auth client has to read from a file rather than being passed one. I'm going to take the approach that is currently implemented and move it around a bit to block any accidental uploads of secrets and use the erb templating but it annoyingly has to be saved.

app/controllers/brexit_feedback_controller.rb
JonathanHallam: So as it stands the auth method is being called for every request. I've moved things around a bit (pr up soon) so that there's one static file that holds a json.erb template and a method that writes to a new file in the root directory. I'm not too bothered by the fact that the authorisation isn't saved because the use case for this is one person running it once a week. If this changes I'd be happy to look into making this a little smoother. Do you know if there's an elegant way to delete a file after a page has been rendered?

app/controllers/brexit_feedback_controller.rb
steventux: https://www.rubydoc.info/github/google/google-api-ruby-client/Google%2FAPIClient%2FClientSecrets.load takes a filename as an argument, this means you can do the ERB template dance and write the file out (as a Tempfile perhaps as it will contain real secrets) and pass the path of the tempfile to `Google::APIClient::ClientSecrets.load(filename)`

app/controllers/brexit_feedback_controller.rb
steventux: I think reading and writing a file for every request is a limited approach, additionally deleting it will only add to the I/O.
It might be better to extend how `Google::APIClient::ClientSecrets.load` works to do what we want here.
If you [look at the source](https://github.com/googleapis/google-api-ruby-client/blob/master/lib/google/api_client/client_secrets.rb#L73) for `Google::APIClient::ClientSecrets.load`  it returns the result of `data = File.open(filename, 'r') { |file| JSON.load(file.read) }` We could extend this class it a load method of our own and make the process much tidier.

client_secrets.json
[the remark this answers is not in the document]
JonathanHallam: 👍

app/models/support/requests/brexit_feedback_request.rb
JonathanHallam: 👍

app/controllers/brexit_feedback_controller.rb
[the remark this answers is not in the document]
JonathanHallam: 👍

app/models/support/requests/brexit_slug_fetcher.rb
```diff
@@ -0,0 +1,38 @@
...
+      def get_google_sheets_slugs
+        config = YAML.load_file("config/google_config.yaml")
+
+        spreadsheet_id = config["spreadsheet_id_1"]
+        range = 'List of documents with facets!A2:A'
+        response = @service.get_sheet_values(spreadsheet_id, range).values.flatten
+
+        spreadsheet_id = config["spreadsheet_id_2"]
+        range = 'Brexit docs!A4:A'
+        response2 = @service.get_sheet_values(spreadsheet_id, range).values.flatten
```
steventux: This is usually a good indicator that you're not naming variables very descriptively. These are 2 different API responses from reading 2 different sheets, is there a name which indicates which sheet is which? Also if you reversed the read order the var names wouldn't make sense...

config/google_config.yaml
```diff
@@ -0,0 +1,3 @@
+---
+  spreadsheet_id_1: 1bFSDYFT5fBpDQTvAeqw4j7QhYXTnFmDuGCLGDwx-wYk
```
steventux: As mentioned, these might be better with descriptive names which help the maintainer understand the contents.

spec/models/support/requests/brexit_slug_fetcher_spec.rb
```diff
@@ -0,0 +1 @@
+
```
steventux: Very minimal XD

`BrexitSlugFetcher` is complex enough code that it needs test coverage though.

spec/services/client_secrets_loader_spec.rb
```diff
@@ -0,0 +1,24 @@
+describe ClientSecretsLoader do
+  context "everything's fine" do
+    it "loads the secrets" do
+      ENV["google_client_id"] = "totaly_real_client_id"
```
steventux: A trick with ENV vars and tests is to always reset them after the test otherwise you pollute the rest of your test suite.
This can often lead to unintentional fails and passes.
https://thoughtbot.com/blog/testing-and-environment-variables#override-environment-variables-on-a-per-test-basis

JonathanHallam: yeah found this out the hard way! Is deleting them on lines 8 and 9 not enough? is there something I'm missing here?

spec/models/support/requests/brexit_slug_fetcher_spec.rb
JonathanHallam: oops, meant to delete this file! will have a think about ways I can test it

spec/services/client_secrets_loader_spec.rb
steventux: Sorry my bad, I missed the deleting bits.
It might be clearer to delete and then set the ENV vars in a before block.

JonathanHallam: 👍

JonathanHallam:
Closing because this has been superseded by https://github.com/alphagov/support/pull/606

PR #610, opened 2019-04-26
Don't expand params when loading paths

thomasleese:
This fixed a bug that was introduced in #606. In that implementation, the `params` variable gets modified with the expanded paths. However this means the pagination, which is automatically generated based on the params, links to the expanded paths rather than the shorted params. Therefore users wouldn't be able to visit second pages when uploading lots of URLs because the same problem as before would occur.

[Trello Card](https://trello.com/c/zfFWlA9r/965-feedex-error-when-viewing-results)

### DIFF
--- a/app/controllers/anonymous_feedback_controller.rb
+++ b/app/controllers/anonymous_feedback_controller.rb
@@ -68,13 +68,19 @@ private
     if params[:path].present?
       params[:paths] = [params[:path]]
     elsif params[:paths] && params[:paths].instance_of?(String)
-      saved_paths = Support::Requests::Anonymous::Paths.find(params[:paths])
-      params[:paths] = saved_paths.try(:paths) || params[:paths].split(',').map(&:strip)
+      params[:paths] = params[:paths].split(',').map(&:strip)
     end
   end
 
+  def paths
+    return [] unless index_params[:paths].present?
+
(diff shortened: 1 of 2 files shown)


app/controllers/anonymous_feedback_controller.rb
```diff
@@ -66,15 +66,21 @@ def index_params
 
   def clean_paths
     if params[:path].present?
-      params[:paths] = [params[:path]]
+      params[:paths] = params[:path]
```
barrucadu: This is a no-op

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -66,15 +66,21 @@ def index_params
...
     elsif params[:paths] && params[:paths].instance_of?(String)
-      saved_paths = Support::Requests::Anonymous::Paths.find(params[:paths])
-      params[:paths] = saved_paths.try(:paths) || params[:paths].split(',').map(&:strip)
+      params[:paths] = params[:paths].split(',').map(&:strip)
```
barrucadu: The `elsif` returns a list, the `if` returns a path.  Is that right?

thomasleese: Yep, good point, I've changed that back to the `if` returns a list. I'm not sure if we actually need to support the old `path` param anymore, but not confident enough to remove it.

PR #616, opened 2019-05-08
Update business finder request to use facet values from Publishing API

AgaDufrat:
Facet values are currently hardcoded in the EU Exit business readiness request from and therefore have to be manually changed when the facet value is updated in the finder.

This populates a support form with values obtained from the Publishing API.

Dependent on https://github.com/alphagov/govuk-puppet/pull/9080 

Test by checking if options for _Is the content about selling goods in the UK, importing or doing business abroad?_ question are in the ground form (eg. _sell_, not _selling_)

[Trello card](https://trello.com/c/HPOb08N2/69-change-business-activity-title-to-organisation-activity-in-zendesk-ticket-generated-by-support-form)

### DIFF
--- a/app/models/support/requests/eu_exit_business_readiness_request.rb
+++ b/app/models/support/requests/eu_exit_business_readiness_request.rb
@@ -3,10 +3,9 @@ require 'active_support/core_ext'
 module Support
   module Requests
     class EuExitBusinessReadinessRequest < Request
-      attr_accessor :type, :url, :explanation, :sector, :business_activity,
-                    :employing_eu_citizens, :personal_data,
-                    :intellectual_property, :funding_schemes,
-                    :public_sector_procurement, :pinned_content
+      attr_accessor :type, :url, :explanation, :sector, :organisation_activity,
+        :employing_eu_citizens, :personal_data, :intellectual_property,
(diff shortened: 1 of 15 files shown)


AgaDufrat:
Updated questions according to [Sharon's instructions](https://docs.google.com/document/d/1ycfjSZ93kZ3ZTy_pRGNK5n24M6Zo2TEyJtlYEENsYfo)

app/models/support/requests/eu_exit_business_readiness_request.rb
```diff
@@ -122,6 +52,16 @@ def self.label
       def self.description
         "Request to add content, update content tags or remove content from the EU Exit business readiness finder."
       end
+
+    private
+
+      def raw_facet_group
+        RemoteFacetGroupService.new.find(EU_EXIT_BUSINESS_FINDER_CONTENT_ID)
```
steventux: Not a blocker for this PR as this method is only every called once per support request but this could be memoized so that we don't inadvertently re-request the facet group. eg.

```ruby
@facet_group ||= RemoteFacetGroupService.new.find(EU_EXIT_BUSINESS_FINDER_CONTENT_ID)
```

PR #665, opened 2019-09-18
Add brexit checker support form

emmabeynon:
For https://trello.com/c/bgtCMkNP/243-zendesk-content-change-request-form

This creates a request form for departments to request a change to an existing action or creating a new action for the Get Ready For Brexit Checker.  Once submitted, a request is made to the Zendesk API to create a ticket.

brexit_check_form

### DIFF
--- a/app/controllers/brexit_checker_requests_controller.rb
+++ b/app/controllers/brexit_checker_requests_controller.rb
@@ -26,6 +26,7 @@ protected
             :new_action_lead_time,
             :new_action_deadline,
             :new_action_comments,
+            :new_action_priority,
             requester_attributes: %i[email name collaborator_emails],
           )
           .to_h
--- a/app/controllers/brexit_checker_requests_controller.rb
+++ b/app/controllers/brexit_checker_requests_controller.rb
@@ -4,4 +4,22 @@ protected
   def new_request
     @get_ready_for_brexit_checker_request = Support::Requests::BrexitCheckerRequest.new
   end
+
+  def zendesk_ticket_class
+    Zendesk::Ticket::BrexitCheckerTicket
+  end
+
+  def parse_request_from_params
+    Support::Requests::BrexitCheckerRequest.new(brexit_checker_params)
+  end
+
+  def brexit_checker_params
+    params.require(:support_requests_brexit_checker_request)
+          .permit(
+            :action_to_change,
+            :description_of_change,
+            requester_attributes: %i[email name collaborator_emails],
+          )
+          .to_h
+  end
 end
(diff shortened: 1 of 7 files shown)


spec/features/get_ready_for_brexit_checker_request_spec.rb
```diff
@@ -11,7 +11,21 @@
...
+      "requester" => hash_including("name" => "John Smith", "email" => "john.smith@agency.gov.uk"),
+      "comment" => {
+        "body" =>
+"[Change title]
+Action title
+
+[Change description]
+Action description"
+      }
+    )
```
theseanything: I think this needs indentation.

app/views/get_ready_for_brexit_checker_requests/_request_details.html.erb
```diff
@@ -29,4 +29,57 @@
...
+    :add_lead_time,
+    label: "Lead time",
+    hint: "How long does the process take, if applicable?",
+    ) %>
+
+  <%= f.input(
+    :add_deadline,
+    label: "Deadline",
+    hint: "Does the user need to act before a certain date?",
+    ) %>
```
theseanything: Do we need an input for priority?

spec/features/get_ready_for_brexit_checker_request_spec.rb
emmabeynon: I had originally thought so too, but it seems if you indent then it affects the API request that's made, so you get extra whitespace, like this:
`"comment"=>{"body"=>"[Action to change]\n          Action title\n\n          [Description of change]\n          Action description"`

When it should be this:
`"comment":{"body":"[Action to change]\nAction title\n\n[Description of change]\nAction description"`

There is precedent for this in the app, for example: https://github.com/alphagov/support/blob/master/spec/features/eu_exit_business_readiness_request_spec.rb#L21

app/views/get_ready_for_brexit_checker_requests/_request_details.html.erb
emmabeynon: No, the default in Zendesk is 'normal' and we don't have a requirement to change it to another level.

theseanything: Ah sorry, I mean for the new action.

emmabeynon: Ah I see - will check with Ben.

emmabeynon: Priority field has been added 👍

PR #672, opened 2019-09-26
Error when exporting csv with large number of URL's

MahmudH:
When users click 'Export as CSV' , they get the error '502 Bad Gateway' if the path has a large amount of paths present.

To solve this problem, instead of POST'ing the large number of URL's we have added a path_id we get from REDIS and POST the ID instead of the URL's.

Alan Gabbianelli <alan.gabbianelli@digital.cabinet-office.gov.uk>


app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -48,8 +51,22 @@ def clean_paths
     end
   end
 
+  def saved_paths
+    @saved_paths ||= anonymous_feedback_params[:paths].present? ? Support::Requests::Anonymous::Paths.find(anonymous_feedback_params[:paths].first) : nil
```
cbaines: I'm struggling to understand what the `.first` is doing here. I can see a similar occurrence in the code prior to these changes, but I can't really see what it's doing there either. Any ideas?

thomasleese: The `clean_paths` method above turns it into an array. For example, if the value is `path1,path2` it'll get parsed into `%w(path1 path2)`.

This allows the controller to support the old style URLs (with the paths given directly in the URL parameters) and the new stlye URLs (where the paths is an ID of the saved paths).

cbaines: Hmm, I don't think I understand the old and new style URLs. So to clarify, the `paths` query parameter is either used for a list of paths as you describe, or as a single 10 character identifier, and the code here is assuming that it's the latter?

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -173,6 +173,20 @@
     end
   end
 
+  describe "#path_id" do
+    include GdsApi::TestHelpers::SupportApi
+
+    it "is true when path_id is present" do
```
cbaines: What does true mean here? I'd expect the test to say `.to be_true` from the above line.

app/views/anonymous_feedback/index.html.erb
```diff
@@ -11,7 +11,7 @@
                                scopes: @filtered_by) %>
     <%= hidden_field_tag :from, @dates.from if @dates.from.present? %>
     <%= hidden_field_tag :to, @dates.to if @dates.to.present? %>
-    <%= hidden_field_tag :paths, @filtered_by.paths if @filtered_by.paths.present? %>
+    <%= hidden_field_tag :paths, @filtered_by.path_id if @filtered_by.path_id.present? %>
```
cbaines: If the contents of this hidden field has changed, could the name change as well?

A field name like `path_id` might make more sense.

app/controllers/anonymous_feedback/export_requests_controller.rb
thomasleese: Yep, that's correct.

app/views/anonymous_feedback/index.html.erb
MahmudH: Yes you're right, I will make that correction. Thanks

spec/presenters/scope_filters_presenter_spec.rb
MahmudH: I forgot to change the title of the test when I copied over another test, thanks for pointing that out. Will change to check that the path_id is present 👍

app/views/anonymous_feedback/index.html.erb
cbaines: Now I'm looking at this again, I think `path_id` is definitely better than `paths`, but something like `path_set_id` might be even clearer, as the id doesn't identify a single path, but a set of paths.

MahmudH: I've changed `path_id` to `path_set_id`

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -36,20 +37,25 @@ def export_request_params
...
-      params[:paths] = params[:paths].split(",").map(&:strip)
-    end
+  def saved_paths
+    @saved_paths ||= Support::Requests::Anonymous::Paths.find(anonymous_feedback_params[:path_set_id])
+  end
+
+  def paths
+    return [] if anonymous_feedback_params[:path_set_id].blank?
+
+    saved_paths.try(:paths)
```
thomasleese: Do we need to use `try` or can be just do `.paths`? I'd imagine now that we don't support `params[:paths]`, we'd want it to error if the `saved_paths` doesn't exist.

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -36,20 +37,25 @@ def export_request_params
...
   end
 
   def scope_filters
-    @scope_filters ||= ScopeFiltersPresenter.new(paths: anonymous_feedback_params[:paths], organisation_slug: anonymous_feedback_params[:organisation])
+    @scope_filters ||= ScopeFiltersPresenter.new(
+      paths: paths,
+      path_set_id: saved_paths.try(:id),
```
thomasleese: Same here with the `try`.

app/controllers/anonymous_feedback/export_requests_controller.rb
cbaines: What was the resolution to this?

PR #675, opened 2019-10-01
Auto-fix linting issues

thomasleese:
These were missed in https://github.com/alphagov/support/pull/669.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -52,7 +52,7 @@ Rails.application.configure do
   config.log_level = :info
 
   # Prepend all log lines with the following tags.
-  config.log_tags = [ :request_id ]
+  config.log_tags = [:request_id]
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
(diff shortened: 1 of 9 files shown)


config/initializers/inflections.rb
```diff
@@ -10,7 +10,7 @@
 #   inflect.uncountable %w( fish sheep )
 # end
 ActiveSupport::Inflector.inflections(:en) do |inflect|
-  inflect.uncountable %w( feedback )
+  inflect.uncountable %w(feedback)
```
AlanGabbianelli: This is inconsistent with `%w[string]` that we use elsewhere. We should probably use one or the other, not a mix of both.

thomasleese: Hmm, I personally use `%w()` rather than square-brackets.

config/environments/production.rb
```diff
@@ -52,7 +52,7 @@
   config.log_level = :info
 
   # Prepend all log lines with the following tags.
-  config.log_tags = [ :request_id ]
+  config.log_tags = [:request_id]
```
AlanGabbianelli: Shouldn't we use %i[request_id] here?

config/initializers/inflections.rb
AlanGabbianelli: That's fine, I don't mind which one, to be honest. I just think it should be consistent :)

PR #687, opened 2019-10-29
Fix support for local docker dev environment

callumknights:
Related: https://github.com/alphagov/link-checker-api/pull/284

Custom alias domains such as http://support.dev.gov.uk are not working
on any apps running Rails 6, due to the addition of "Host Authorisation"
middleware.

Without this fix you get this error when trying to access the support
app in browser.

> Blocked host: support.dev.gov.uk
> To allow requests to support.dev.gov.uk, add the following to your environment configuration:
> config.hosts << support.dev.gov.uk


thomasleese:
I wonder if we can get https://github.com/alphagov/govuk_app_config to do this for us automatically.

PR #724, opened 2020-01-21
Improve the .gitignore

cbaines:


cbaines:
I think the tests are just broken in `master` at the moment, so this will have to wait until they're fixed...

issyl0:
If we rebase this, the tests will pass.

cbaines:
> If we rebase this, the tests will pass.

They worked on the 2nd attempt at least :)

PR #765, opened 2020-04-17
Ensure Feedback Export requests have a valid `path_prefixes` param

edwardkerry:
Recently we refactored feedback CSV exports so that post requests are
sent with a Redis `path_set_id` rather than a large list of URLs.
https://github.com/alphagov/support/pull/672

Whilst it no longer possible with the Support app itself to access the `anonymous_feedback` path without first generating a `path_set_id`, direct links do exist within the GOV.UK Chrome extension and the Content-Data app. This causes the ExportRequestsController to send through a request to the Support-API with an invalid `path_prefixes: scope_filters.paths_for_api`

This PR ensures that when directly accessing this controller with a string URL `path` or `paths` param, clicking the `Export as CSV` button will cause the ExportRequestsController to create a valid Redis `path_set_id` if one is not set. This is in turn used to look up the correct `path_prefixes`

[trello](https://trello.com/c/oy8WqDlD/1909-5-support-export-request-doesnt-send-email)

### DIFF
--- a/app/controllers/anonymous_feedback/export_requests_controller.rb
+++ b/app/controllers/anonymous_feedback/export_requests_controller.rb
@@ -5,6 +5,8 @@ class AnonymousFeedback::ExportRequestsController < AuthorisationController
   def create
     authorize! :read, :anonymous_feedback
 
+    set_path_set_id_param
+
     support_api.create_feedback_export_request(export_request_params)
 
     redirect_to anonymous_feedback_index_path(anonymous_feedback_params),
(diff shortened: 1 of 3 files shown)


app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -26,6 +28,18 @@ def show
 
 private
 
+  def set_path_set_id_param
+    return true if anonymous_feedback_params[:path_set_id].presence
```
thomasleese, line 32: Do we need to return `true` or can we just return on its own? And should it be `.present?` rather than `presence`.

app/controllers/anonymous_feedback/export_requests_controller.rb
```diff
@@ -26,6 +28,18 @@ def show
...
+
+    paths = anonymous_feedback_params[:paths].presence
```
thomasleese, line 34:
```suggestion
    paths = anonymous_feedback_params[:paths].presence
    return unless paths
```

Then we don't need the `if` below and it follows the pattern of the line above.

edwardkerry: 👍

app/controllers/anonymous_feedback/export_requests_controller.rb
edwardkerry: whoops, thanks!

PR #820, opened 2020-09-10
Upgrade ruby 2.7

benbaumann95:
- Tested locally with ruby 2.7.1 and 2.6.6
- Upon upgrading to 2.7.1, there was a whirlwind of deprecation warnings and incompatibility issues with formtastic / formatstic-bootstrap gem as explained in https://github.com/alphagov/support/commit/6c6d60bb6064cb412e6d672d3ca44e05ba33330b
- Re-wrote all formtastic forms natively. No cosmetic change but should be far more future-proof and easier to maintain


app/views/content_advice_requests/_request_details.html.erb
```diff
@@ -2,34 +2,78 @@
...
 
-<%= f.inputs name: "Details of your request" do %>
-  <%= f.input :details, as: :text, required: true,
-              label: "Please explain what you would like help with",
-              input_html: { class: "input-md-6", rows: 6, cols: 50, :"aria-required" => true } %>
+  <div class="form-group">
+    <span class="form-label">
+      <%= f.label :details do %>
+        Please explain what you would like help with<abbr title="required">*</abbr>
+      <% end %>
```
cdccollins, line 23: In the app I was working on they did this kind of thing another way -

```
      <%= f.label :justifications, "Is this need in proposition for GOV.UK? You can tick more than one.<br/><span class='optional'>(optional)</span>".html_safe %>
```

I don't know if one way is preferable

app/views/requests/new.html.erb
```diff
@@ -3,8 +3,8 @@
 
 <% breadcrumb :request, @request %>
 
-<%= semantic_form_for @request, url: { action: "create" }, html: { novalidate: false } do |f| %>
+<%= form_for @request, url: { action: "create" }, html: { novalidate: false } do |f| %>
   <%= render partial: "request_details", locals: { f: f } %>
 
-  <%= f.action :submit, label: "Submit", button_html: { class: "btn btn-success" } %>
+  <%= f.submit value: "Submit", class: "btn btn-success" %>
```
cdccollins, line 9: I think you can just do `f.submit, class: "whatever"` or maybe `f.submit :submit, class: "whatever`

cdccollins (approved):
Just a few comments about possible other ways of doing things. 

I would also like someone else to take a look at this, just because there is a lot and I may have missed something

PR #871, opened 2021-04-16
Update campaign form

GDSNewt:
We want to update the campaigns application form in the Support app to reflect changes in the way campaigns are applied for.

To this end we want to remove the existing text from the current form 

 • A campaign platform site is a one page website on campaign.gov.uk
 • Microsite websites are not built on GOV.UK and may or may not have a gov.uk url. Microsite websites fall under Cabinet 
    Office digital service expenditure therefore GDS spend approval must be sought following approval from GCS to 
    develop a campaign microsite.*

We also want to remove any reference to microsites. 

**BEFORE**
Screenshot 2021-04-16 at 13 35 05

**AFTER**
Screenshot 2021-04-27 at 10 02 12

[Trello](https://trello.com/c/C7maXqA2/324-non-covid-update-campaigns-application-form-in-support-app)

### DIFF
--- a/app/views/campaign_requests/_request_details.html.erb
+++ b/app/views/campaign_requests/_request_details.html.erb
@@ -3,15 +3,9 @@
 </div>
 
 <p>
-  Use this form for : Campaign Platform site or microsite
-<ul>
-  <li>A campaign platform site is a one page website on campaign.gov.uk</li>
-  <li>Microsite websites are not built on GOV.UK and may or may not have
-    a gov.uk url.
-    Microsite websites <%= link_to 'fall under Cabinet Office digital service expenditure', 'https://www.google.com/url?q=https://www.gov.uk/service-manual/agile-delivery/spend-controls-check-if-you-need-approval-to-spend-money-on-a-service%23when-your-service-is-considered-a-digital-service&sa=D&ust=1518444538536000&usg=AFQjCNFvnZ81hn7cqJnGHgcgQ0V0bdwCBQ'%>
-    therefore GDS spend approval must be sought following approval from GCS to develop a campaign microsite.
-  </li>
-</ul>
+  Use this form for: Wordpress “single page” Campaign Platform site, or bespoke “multi page” site
+</p>
+<p>
   For either type of site, please answer the questions below.
 </p>
 
@@ -24,11 +18,11 @@
     <div class="form-group">
       <span class="form-label control-label">
         <%= r.label :type_of_site do %>
-          Are you applying for the campaign platform or a bespoke microsite?<abbr title="required">*</abbr>
+          Are you applying for the single page campaign platform or a bespoke multi page site?<abbr title="required">*</abbr>
         <% end %>
       </span>
       <span class="form-wrapper">
-        <%= r.collection_radio_buttons :type_of_site, ["Campaign platform", "Bespoke microsite"], :itself, :itself, required: true, aria: { required: true } do |builder| %>
+        <%= r.collection_radio_buttons :type_of_site, ["Single page campaign platform site", "Multi page site"], :itself, :itself, required: true, aria: { required: true } do |builder| %>
           <div class="radio">
             <%= builder.label(class: "custom-control-label") do %>
               <%= builder.radio_button(class: "custom-control-input") %>
@@ -209,7 +203,7 @@
     </div>
     <p class="help-block">
       If applying for Campaigns Platform it is in the form of xxxxx.campaign.gov.uk, or otherwise
-      if a microsite, standard format is  in the form of xxxxxx.gov.uk.
+      if a bespoke multi page site, standard format is in the form of xxxxxx.gov.uk.
     </p>
 
     <div class="form-group">
--- a/app/models/support/gds/campaign.rb
+++ b/app/models/support/gds/campaign.rb
@@ -40,7 +40,7 @@ module Support
                 presence: true
 
       validates :has_read_guidance_confirmation, :has_read_oasis_guidance_confirmation, acceptance: { allow_nil: false }
-      validates :type_of_site, inclusion: { in: ["Campaign platform", "Bespoke microsite"] }
+      validates :type_of_site, inclusion: { in: ["Single page campaign platform site", "Multi page site"] }
 
       validates_date :start_date, on_or_after: :today
       validates_date :end_date, after: :start_date
--- a/app/models/zendesk/ticket/campaign_request_ticket.rb
+++ b/app/models/zendesk/ticket/campaign_request_ticket.rb
@@ -16,7 +16,7 @@ module Zendesk
           Zendesk::LabelledSnippet.new(
             on: @request.campaign,
             field: :type_of_site,
-            label: "Are you applying for the campaign platform or a bespoke microsite?",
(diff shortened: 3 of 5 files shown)


app/views/campaign_requests/_request_details.html.erb
```diff
@@ -3,15 +3,8 @@
...
-  Use this form for : Campaign Platform site or microsite
-<ul>
-  <li>A campaign platform site is a one page website on campaign.gov.uk</li>
-  <li>Microsite websites are not built on GOV.UK and may or may not have
-    a gov.uk url.
-    Microsite websites <%= link_to 'fall under Cabinet Office digital service expenditure', 'https://www.google.com/url?q=https://www.gov.uk/service-manual/agile-delivery/spend-controls-check-if-you-need-approval-to-spend-money-on-a-service%23when-your-service-is-considered-a-digital-service&sa=D&ust=1518444538536000&usg=AFQjCNFvnZ81hn7cqJnGHgcgQ0V0bdwCBQ'%>
-    therefore GDS spend approval must be sought following approval from GCS to develop a campaign microsite.
-  </li>
-</ul>
+  Use this form for : Campaign Platform site or a multi-page campaign <br>
```
Rosa-Fox, line 6: I know it was already there, but would you be able to remove the space between `for` and the `:`?

I think it might be worth going back to product by commenting on the Trello ticket or messaging Lisa to find out if they have any preference for the content... I know I am not a content designer 😅 but I think something along the lines of:

```
<p>
  Use this form for a campaign platform site or a multi-page campaign.
</p>

<p>
  For either type of site, please answer the questions below.
</p>
```

would work better. I also think we should remove the `br` tags. It looks a bit tidier. Semantically they could be in the same paragraph but I think it also works having them separate. Having them separate stops the text wrapping around on a smaller screen which reads a bit better.

If it will take too long for the content question to be answered then:

```
<p>
  Use this form for: Campaign Platform site or a multi-page campaign.
</p>
<p>
  For either type of site, please answer the questions below.
</p>
```

and we can update it at a later date.

Rosa-Fox (commented):
@GDSNewt I ran this locally and it functions well :) I have some comments but I think they are content/product decision influenced rather than a problem with the implementation itself.

I think we should find out how `Campaign platform` and `multi-page campaign` should be capitalised and then make that consistent for both the user facing text and also the form value that is sent to the backend. 

I also noticed that `microsite` is used further down the page:
- `If applying for Campaigns Platform it is in the form of xxxxx.campaign.gov.uk, or otherwise if a microsite, standard format is in the form of xxxxxx.gov.uk.` 
- `microsites@cabinetoffice.gov.uk`

I couldn't see any requirements in the ticket that explain what should be done with these, so I think you will need to go back and ask. I also left a comment with some more questions to ask the other disciplines.

Let me know if you have any questions or need any help getting hold of people.

app/views/campaign_requests/_request_details.html.erb
GDSNewt: Thanks for the review Rosa, now using 2 <p> as you recommended. I used br initially as I thought 2 <p> seemed heavy handed, but I hadn't thought of the wrapping issue 😅 .

GDSNewt:
> @GDSNewt I ran this locally and it functions well :) I have some comments but I think they are content/product decision influenced rather than a problem with the implementation itself.
> 
> I think we should find out how `Campaign platform` and `multi-page campaign` should be capitalised and then make that consistent for both the user facing text and also the form value that is sent to the backend.
> 
> I also noticed that `microsite` is used further down the page:
> 
> * `If applying for Campaigns Platform it is in the form of xxxxx.campaign.gov.uk, or otherwise if a microsite, standard format is in the form of xxxxxx.gov.uk.`
> * `microsites@cabinetoffice.gov.uk`
> 
> I couldn't see any requirements in the ticket that explain what should be done with these, so I think you will need to go back and ask. I also left a comment with some more questions to ask the other disciplines.
> 
> Let me know if you have any questions or need any help getting hold of people.



> @GDSNewt I ran this locally and it functions well :) I have some comments but I think they are content/product decision influenced rather than a problem with the implementation itself.
> 
> I think we should find out how `Campaign platform` and `multi-page campaign` should be capitalised and then make that consistent for both the user facing text and also the form value that is sent to the backend.
> 
> I also noticed that `microsite` is used further down the page:
> 
> * `If applying for Campaigns Platform it is in the form of xxxxx.campaign.gov.uk, or otherwise if a microsite, standard format is in the form of xxxxxx.gov.uk.`
> * `microsites@cabinetoffice.gov.uk`
> 
> I couldn't see any requirements in the ticket that explain what should be done with these, so I think you will need to go back and ask. I also left a comment with some more questions to ask the other disciplines.
> 
> Let me know if you have any questions or need any help getting hold of people.

There should only be one reference to microsite left now, that being the email address - apparently that is still in use. The capitalisation is a bit inconsistent but what the content people want. 

> @GDSNewt I ran this locally and it functions well :) I have some comments but I think they are content/product decision influenced rather than a problem with the implementation itself.
> 
> I think we should find out how `Campaign platform` and `multi-page campaign` should be capitalised and then make that consistent for both the user facing text and also the form value that is sent to the backend.
> 
> I also noticed that `microsite` is used further down the page:
> 
> * `If applying for Campaigns Platform it is in the form of xxxxx.campaign.gov.uk, or otherwise if a microsite, standard format is in the form of xxxxxx.gov.uk.`
> * `microsites@cabinetoffice.gov.uk`
> 
> I couldn't see any requirements in the ticket that explain what should be done with these, so I think you will need to go back and ask. I also left a comment with some more questions to ask the other disciplines.
> 
> Let me know if you have any questions or need any help getting hold of people.

There should only be one reference to microsite left now, that being the email address - apparently that is still in use. The capitalisation is a bit inconsistent but it seems that's what the content people want. I think all the backend stuff, i'e anything that isn't a label is consistent with the checkbox descriptions.

app/views/campaign_requests/_request_details.html.erb
```diff
@@ -24,11 +18,11 @@
     <div class="form-group">
       <span class="form-label control-label">
         <%= r.label :type_of_site do %>
-          Are you applying for the campaign platform or a bespoke microsite?<abbr title="required">*</abbr>
+          Are you applying for the single page campaign platform or a bespoke multi page site ?*<abbr title="required">*</abbr>
```
Rosa-Fox, line 21: I would remove the space between `site` and the `?` in `multi page site ?`... I know it was like that in the copy but...

Rosa-Fox (changes requested):
Thanks! I had one comment. Also the copy is inconsistent with the use of hyphens for `multi-page`. It is different on nearly each line (I know it was given to you like this). I think we should take out all of the hyphens from every `multi-page` as `single page` doesn't have them, just to make it look more consistent. Then we can merge!

app/views/campaign_requests/_request_details.html.erb
GDSNewt: Done 😸

Rosa-Fox (approved):

app/views/campaign_requests/_request_details.html.erb
Rosa-Fox: thanks :)

PR #970, opened 2022-05-16
Add all Publishing services to the technical fault report page

davidgisbey:
## Description

Users can report technical faults on publishing services via the technical fault report page. At the moment, there is an incomplete list of services which can lead to confusion for users and colleagues when triaging feedback.

This commit adds in missing Publishing services. It also adds in a `do_not_know` option.

It felt wrong to add this to the list the `all` method in the UserFacingComponents module, as it's not a user facing service.

I've chosen instead to update the technical_fault_report class to add it as an extra option and ensure that validations are not
triggered if it selected by the user.

It might be a bit overkill, but it allows the UserFacingComponents class to pass the correct user services back and increases reusability if this application is ever extended.

## Trello card 

https://trello.com/c/WG2MPjSd/401-support-form-doesnt-include-all-the-publishing-apps


app/models/support/gds/user_facing_components.rb
```diff
@@ -23,6 +23,14 @@ def component_options
...
             { name: "Content Publisher (beta)", id: "content_publisher" },
             { name: "data.gov.uk", id: "datagovuk" },
+            { name: "Collections Publisher", id: "collections_publisher" },
+            { name: "Contacts Admin", id: "contacts_admin" },
+            { name: "Content Tagger", id: "content_tagger" },
+            { name: "Local Links Manager", id: "local_links_manager" },
+            { name: "Manuals Publisher", id: "manuals_publisher" },
+            { name: "Service Manual Publisher", id: "service_manual_publisher" },
+            { name: "Short URL Manager", id: "short_url_manager" },
+            { name: "Maslow", id: "maslow" },
```
davidgisbey, line 33: Just double checking. Is this a complete list?

beccapearce: Looks good, Not sure if we need content data and imminence as well? But maybe we can keep an eye out when we're on support to see if we get any requests for them in the don't know option. 
I'm curious also if there is an order to this list (e.g. most commonly used?) If there isn't might be nice to have them alphabetically.

davidgisbey: Good shout! I'll add them in! (Ngl i've literally never heard of Imminence haha)

In terms of ordering. I'd assumed the most commonly added applications were at the top. Having said that i think it'd be pretty nice if they were alphabetised. I'll update it to be ordered 👍 

Thanks for the feedback Becca!

beccapearce (approved):

PR #1079, opened 2023-01-18
Include app name in the rails assets path

samsimpson1:
Fixes assets not being served correctly on k8s

See also: https://github.com/alphagov/manuals-publisher/pull/2004

https://trello.com/c/VuxjcRPI/1096-user-journey-make-a-support-request

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -30,5 +30,7 @@ module Support
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+
+    config.assets.prefix = "/assets/support"
   end
 end


config/application.rb
```diff
@@ -30,5 +30,7 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+
+    config.assets.prefix = "/assets/collections-publisher"
```
sengi, line 34: Copy-pasto?

samsimpson1: Its a good job one of us is paying attention. Fixed

sengi (approved):

PR #1086, opened 2023-01-31
Remove unused class, drop unmaintained dependency & update Ruby to 3.1.2

ChrisBAshton:
This was supposed to be a quick bit of cleanup (removing a bit of code we've identified as no longer being used), but that caused a failing GitHub Brakeman check on our Ruby version. Evidently, we're not allowed to make code changes without moving to a supported Ruby version.

So I updated the Ruby version to the next major version, which was incompatible with one of our dependencies, so I updated that too. Finally, I've bumped to the next Ruby version again, because why not?


deborahchua (approved):

KludgeKML (approved):
Looks good to me.

PR #1087, opened 2023-02-02
[DO NOT MERGE] Changes to content requests form in publisher

DilwoarH:
## Note

- All form updates have been done
- The only thing remaining to do is the mapping on Zendesk.

```
- Custom “title” field
- Custom “reason for change” field
- Custom “subject area” field
- Custom “needs Welsh” field, based on “Welsh URL” field being filled
- “What’s the new or different information”, in Zendesk body
- “Why is change needed”, in Zendesk body
- Custom “Deadline” field (this may already happen)
- Custom “Not before” field (this may already happen)
- “Reason for deadline”, in Zendesk body
- “Who needs a copy of this request”, put into Zendesk cc field (this already happens)
```

Details of change request are added to the following doc:
https://docs.google.com/document/d/1s0OoBexbHXnDIx554EyudsdAYpAKI7C1p4VFBvqkW9Q/edit

## What

Content designers want us to improve [the GOV.UK content request form](https://support.publishing.service.gov.uk/content_change_request/new), which departments and agencies fill in when they want us to change their mainstream content. (This is the form which generates the Zendesk tickets.)

## Why

The form is not fit for purpose: it does not ask departments for a lot of the information that we require nor does it give guidance in the right places.

### Before

screenshot-support dev gov uk-2023 02 02-12_26_44

### After

screencapture-support-dev-gov-uk-content-change-request-new-2023-04-12-16_36_09

New Trello Card: https://trello.com/c/aPOXVkPS/3117-complete-changes-to-content-requests-form-in-support-app-3

~Previous Trello Card: https://trello.com/c/SH6tA5pV/472-changes-to-content-requests-form-in-publisher~

---

### DIFF
--- a/app/controllers/content_change_requests_controller.rb
+++ b/app/controllers/content_change_requests_controller.rb
@@ -23,7 +23,7 @@ protected
       :url,
       :related_urls,
       requester_attributes: %i[email name collaborator_emails],
-      time_constraint_attributes: %i[not_before_date needed_by_date time_constraint_reason],
+      time_constraint_attributes: %i[not_before_date needed_by_date time_constraint_reason needed_by_time not_before_time],
     ).to_h
   end
 end
--- a/app/controllers/content_change_requests_controller.rb
+++ b/app/controllers/content_change_requests_controller.rb
@@ -17,6 +17,7 @@ protected
     params.require(:support_requests_content_change_request).permit(
       :title,
       :reason_for_change,
+      :subject_area,
       :details_of_change,
       :url,
       :related_urls,
--- a/app/controllers/content_change_requests_controller.rb
+++ b/app/controllers/content_change_requests_controller.rb
@@ -19,6 +19,7 @@ protected
       :reason_for_change,
       :subject_area,
       :details_of_change,
+      :why_is_change_needed,
       :url,
       :related_urls,
       requester_attributes: %i[email name collaborator_emails],
--- a/app/controllers/content_change_requests_controller.rb
+++ b/app/controllers/content_change_requests_controller.rb
@@ -16,6 +16,7 @@ protected
   def content_change_request_params
     params.require(:support_requests_content_change_request).permit(
       :title,
+      :reason_for_change,
       :details_of_change,
       :url,
       :related_urls,
--- a/app/controllers/content_change_requests_controller.rb
+++ b/app/controllers/content_change_requests_controller.rb
@@ -19,7 +19,6 @@ protected
       :reason_for_change,
       :subject_area,
       :details_of_change,
-      :why_is_change_needed,
       :url,
       :related_urls,
       requester_attributes: %i[email name collaborator_emails],
--- a/app/models/support/requests/content_change_request.rb
+++ b/app/models/support/requests/content_change_request.rb
@@ -16,7 +16,7 @@ module Support
       end
 
       def self.label
-        "Content changes and new content requests"
+        "Request a content change or new content on GOV.UK"
       end
 
       def self.description
--- a/app/models/support/requests/content_change_request.rb
+++ b/app/models/support/requests/content_change_request.rb
@@ -5,7 +5,7 @@ module Support
     class ContentChangeRequest < Request
       include WithTimeConstraint
 
-      attr_accessor :title, :details_of_change, :url, :related_urls
+      attr_accessor :title, :reason_for_change, :details_of_change, :url, :related_urls
 
       validates :details_of_change, presence: true
 
--- a/app/models/support/requests/content_change_request.rb
+++ b/app/models/support/requests/content_change_request.rb
@@ -5,7 +5,7 @@ module Support
     class ContentChangeRequest < Request
       include WithTimeConstraint
 
-      attr_accessor :title, :reason_for_change, :details_of_change, :url, :related_urls
+      attr_accessor :title, :reason_for_change, :subject_area, :details_of_change, :url, :related_urls
 
       validates :details_of_change, presence: true
 
--- a/app/models/support/requests/content_change_request.rb
+++ b/app/models/support/requests/content_change_request.rb
@@ -5,7 +5,7 @@ module Support
     class ContentChangeRequest < Request
       include WithTimeConstraint
 
-      attr_accessor :title, :reason_for_change, :subject_area, :details_of_change, :url, :related_urls
+      attr_accessor :title, :reason_for_change, :subject_area, :details_of_change, :why_is_change_needed, :url, :related_urls
 
       validates :details_of_change, presence: true
 
--- a/app/models/support/requests/content_change_request.rb
+++ b/app/models/support/requests/content_change_request.rb
@@ -5,7 +5,7 @@ module Support
     class ContentChangeRequest < Request
       include WithTimeConstraint
 
-      attr_accessor :title, :reason_for_change, :subject_area, :details_of_change, :why_is_change_needed, :url, :related_urls
+      attr_accessor :title, :reason_for_change, :subject_area, :details_of_change, :url, :related_urls
 
       validates :details_of_change, presence: true
 
--- a/app/models/support/requests/time_constraint.rb
+++ b/app/models/support/requests/time_constraint.rb
@@ -10,7 +10,8 @@ module Support
 
       validates_date :needed_by_date, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, allow_blank: true, on_or_after: :today
-      validates_date :not_before_date, on_or_before: :needed_by_date, unless: proc { |c| c.needed_by_date.blank? || c.not_before_date.blank? }
+      validates_date :not_before_date, on_or_before: :needed_by_date, unless: proc { |c| c.needed_by_date.blank? || c.not_before_date.blank? },
+                                       message: "'Must not be published before' date cannot be after Deadline"
 
       validates_time :not_before_time, allow_blank: true
       validates_time :needed_by_time, allow_blank: true
--- a/app/models/support/requests/time_constraint.rb
+++ b/app/models/support/requests/time_constraint.rb
@@ -4,11 +4,27 @@ module Support
   module Requests
     class TimeConstraint
       include ActiveModel::Model
-      attr_accessor :not_before_date, :not_before_time, :needed_by_date, :needed_by_time, :time_constraint_reason
+      attr_accessor :not_before_date, :not_before_time, :needed_by_date, :needed_by_time, :time_constraint_reason, :is_legal_deadline
+
+      validates :is_legal_deadline, inclusion: %w[yes no], allow_blank: true
 
       validates_date :needed_by_date, allow_nil: true, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, allow_nil: true, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, on_or_before: :needed_by_date, unless: proc { |c| c.needed_by_date.blank? || c.not_before_date.blank? }
+
+      validates_time :not_before_time, allow_blank: true
+      validates_time :needed_by_time, allow_blank: true
+
+      validates_time :not_before_time, on_or_after: :now, unless: proc { |c|
+        c.not_before_time.blank? || c.not_before_date != Time.zone.today.strftime("%d-%m-%Y")
+      }
+      validates_time :needed_by_time, on_or_after: :now, unless: proc { |c|
+        c.needed_by_time.blank? || c.needed_by_date != Time.zone.today.strftime("%d-%m-%Y")
+      }
+
+      validates_time :not_before_time, on_or_before: :needed_by_time, unless: proc { |c|
+        [c.needed_by_date, c.needed_by_time, c.not_before_date, c.not_before_time].any?(&:blank?) || c.needed_by_date != c.not_before_date
+      }
     end
   end
 end
--- a/app/models/support/requests/time_constraint.rb
+++ b/app/models/support/requests/time_constraint.rb
@@ -4,7 +4,7 @@ module Support
   module Requests
     class TimeConstraint
       include ActiveModel::Model
-      attr_accessor :not_before_date, :needed_by_date, :time_constraint_reason
+      attr_accessor :not_before_date, :not_before_time, :needed_by_date, :needed_by_time, :time_constraint_reason
 
       validates_date :needed_by_date, allow_nil: true, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, allow_nil: true, allow_blank: true, on_or_after: :today
--- a/app/models/support/requests/time_constraint.rb
+++ b/app/models/support/requests/time_constraint.rb
@@ -8,8 +8,8 @@ module Support
 
       validates :is_legal_deadline, inclusion: %w[yes no], allow_blank: true
 
-      validates_date :needed_by_date, allow_nil: true, allow_blank: true, on_or_after: :today
-      validates_date :not_before_date, allow_nil: true, allow_blank: true, on_or_after: :today
+      validates_date :needed_by_date, allow_blank: true, on_or_after: :today
+      validates_date :not_before_date, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, on_or_before: :needed_by_date, unless: proc { |c| c.needed_by_date.blank? || c.not_before_date.blank? }
 
       validates_time :not_before_time, allow_blank: true
--- a/app/models/zendesk/zendesk_ticket.rb
+++ b/app/models/zendesk/zendesk_ticket.rb
@@ -29,6 +29,18 @@ module Zendesk
       end
     end
 
+    def not_before_time
+      if value?(:time_constraint) && value?(:not_before_time, @request.time_constraint)
+        @request.time_constraint.not_before_time
+      end
+    end
+
+    def needed_by_time
+      if value?(:time_constraint) && value?(:needed_by_time, @request.time_constraint)
+        @request.time_constraint.needed_by_time
+      end
+    end
+
     def tags
       %w[govt_form]
     end
@@ -49,7 +61,9 @@ module Zendesk
       if value?(:time_constraint)
         [
           Zendesk::LabelledSnippet.new(on: self, field: :needed_by_date, label: "Needed by date"),
+          Zendesk::LabelledSnippet.new(on: self, field: :needed_by_time, label: "Needed by time"),
           Zendesk::LabelledSnippet.new(on: self, field: :not_before_date, label: "Not before date"),
+          Zendesk::LabelledSnippet.new(on: self, field: :not_before_time, label: "Not before time"),
           Zendesk::LabelledSnippet.new(on: @request.time_constraint, field: :time_constraint_reason, label: "Reason for time constraint"),
         ]
       else
--- a/app/views/content_change_requests/_request_details.html.erb
+++ b/app/views/content_change_requests/_request_details.html.erb
@@ -60,7 +60,7 @@
     <% end%>
   </span>
   <p class="help-block hint-block--md-6" id="subject-area-hint">
-    If none of the options below are appropriate, please select other.
+    If your request covers multiple subject areas, choose the one you think most relevant.
   </p>
   <span class="form-wrapper">
     <%= f.select :subject_area,
@@ -101,7 +101,7 @@
     <% end %>
   </span>
   <p class="help-block hint-block--md-6" id="url-hint">
-    List URLs, one per line.
+    List URLs, one per line. Include any Welsh URLs where translations are needed.
   </p>
   <span class="form-wrapper">
     <%= f.text_area :url, placeholder: "https://www.gov.uk/", required: true, class: "input-md-6 form-control", rows: 4, cols: 50, aria: {
@@ -113,12 +113,13 @@
   <div class="form-group">
     <span class="form-label">
       <%= f.label :details_of_change do %>
-        What’s the new or different information? <abbr title="required">*</abbr>
+        Tell us about the content that needs to be created, updated or is causing a problem for users? <abbr title="required">*</abbr>
       <% end %>
     </span>
     <div class="help-block hint-block--md-6" id="details-of-change-hint">
       <p>Make sure you:</p>
       <ul>
+        <li>Include evidence of why this is a problem, for example, user feedback or call centre data</li>
         <li>give sufficient detail</li>
         <li>use plain English</li>
         <li>explain any acronyms, jargon or legal language</li>
@@ -127,24 +128,6 @@
       </ul>
 
     <p>If you send us wording, we might not use it. If we do, we’ll need to put it in GOV.UK style.</p>
-
-    <p>To ensure your changes are as easy as possible for Content Designers to understand, please following the following format:</p>
-
-    <pre class="well">
-Ch. CHAPTER NAME
-
-[1]
-
-Under #Entry Requirements
-
-[CHANGE]
-
-This is some content that needs to change.
-
-[TO]
-
-This is some content that has changed.
-    </pre>
   </div>
   <span class="form-wrapper">
     <%= f.text_area :details_of_change, required: true, aria: { required: true }, class: "input-md-6 form-control", rows: 16, cols: 50, aria: {
--- a/app/views/content_change_requests/_request_details.html.erb
+++ b/app/views/content_change_requests/_request_details.html.erb
@@ -55,7 +55,13 @@
 <%= render partial: "support/time_constraint", locals: { f: f } %>
 
 <div class="alert alert-info alert-block">
-  <p>If your request is urgent, please fill in this form and call <%= @primary_contact_details[:urgent_department_content_change][:phone] %> during office hours, or <%= @primary_contact_details[:urgent_mainstream_content_change][:phone] %> out of hours. (It's ONLY urgent when the public or the government is facing immediate and significant financial, legal or physical risk.)</p>
+  <p><strong>If your request is urgent</strong></p>
+  <p>Submit this form and call:</p>
+  <ul>
+    <li><%= @primary_contact_details[:urgent_department_content_change][:phone] %> between 9am and 6pm</li>
+    <li><%= @primary_contact_details[:urgent_mainstream_content_change][:phone] %> at other times</li>
+  </ul>
+  <p>It's urgent only if the public or the government faces immediate and significant financial, legal or physical risk.</p>
 </div>
 
 <%= render partial: "support/collaborators", locals: { f: f } %>
--- a/app/views/content_change_requests/_request_details.html.erb
+++ b/app/views/content_change_requests/_request_details.html.erb
@@ -88,37 +88,49 @@
 
 <fieldset>
   <legend>
-    <span>URL affected</span>
+    <span>Which URLs are affected?</span>
   </legend>
 
   <div class="form-group">
     <span class="form-label">
-      <%= f.label :url, "URL" %>
-    </span>
-    <span class="form-wrapper">
-      <%= f.text_field :url, placeholder: "https://www.gov.uk/", class: "input-md-6 form-control" %>
-    </span>
-  </div>
-
-  <div class="form-group">
-    <span class="form-label">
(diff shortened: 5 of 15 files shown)


spec/features/content_change_requests_spec.rb
```diff
@@ -73,14 +85,17 @@
...
-    click_on "Content changes and new content requests"
+    click_on "Request a content change or new content on GOV.UK"
 
     expect(page).to have_content("You'll get an automated response to confirm we've received your request. We'll then review your request within 2 working days.")
 
     fill_in "Title of request", with: details[:title] unless details[:title].nil?
-    fill_in "Details of the requested change", with: details[:details_of_change]
-    fill_in "URL", with: details[:url]
-    fill_in "Does this affect any other URLs (including any existing Welsh translations that need to be updated)? Put each new URL on a new line.", with: details[:related_urls]
+    select details[:reason_for_request], from: "What’s the reason for the request?" unless details[:reason_for_request].nil?
```
deborahchua, line 93: This should be updated if it's meant to be "change" rather than "request"

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -8,54 +8,184 @@
...
   </span>
 </div>
 
-<fieldset>
-  <legend>
-    <span>URL affected</span>
-  </legend>
+<div class="form-group">
+  <span class="form-label">
+    <%= f.label :reason_for_request, "What’s the reason for the request?" %>
```
deborahchua, line 25: Should this be "What's the reason for the change?"? The `reason_for_request` parameter can also be renamed to `reason_for_change` for consistency.

deborahchua (commented):
Looks good! I've left a couple of queries regarding one of the title fields.

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -20,6 +20,33 @@
...
 
+<div class="form-group">
+  <span class="form-label">
+    <%= f.label :reason, "What's the reason for the request?" %>
+  </span>
+  <p class="help-block hint-block--md-6" id="reason-hint">
+    If none of the options below are appropriate, please select other.
+  </p>
+  <span class="form-wrapper">
+    <%= f.select :reason,
```
ChrisBAshton: Do we not need to update [this test](https://github.com/alphagov/support/blob/7e62c6694cda6bf719284e51aaf2fcba131e09d9/spec/features/content_change_requests_spec.rb#L73-L83) to select a reason?

app/controllers/content_change_requests_controller.rb
```diff
@@ -16,6 +16,7 @@ def parse_request_from_params
   def content_change_request_params
     params.require(:support_requests_content_change_request).permit(
       :title,
+      :reason,
```
ChrisBAshton: I notice all the other user-provided inputs have a corresponding [`LabelledSnippet` in the `ContentChangeRequestTicket` model](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/models/zendesk/ticket/content_change_request_ticket.rb#L18) - do we not need that? Is it because the others are text inputs and this one is a dropdown?

EDIT: I notice we also don't add it for the new `why_is_change_needed` field, and that one is a text input 🤔 

EDIT EDIT: ahh, you add them in `Updates test with new fields`. I think you should fix these up to the commits in which you add the fields - makes it easier to review when every change is self-contained 👍

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -85,62 +85,57 @@
   </span>
 </div>
 
-<fieldset>
-  <legend>
-    <span>Which URLs are affected?</span>
-  </legend>
```
ChrisBAshton, line 19: Nice 👍

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -181,7 +181,11 @@ This is some content that has changed.
 <%= render partial: "support/time_constraint", locals: { f: f } %>
 
 <div class="alert alert-info alert-block">
-  <p>If your request is urgent, please fill in this form and call 07810 750 834 during office hours, or 07827 992603 out of hours. (It's ONLY urgent when the public or the government is facing immediate and significant financial, legal or physical risk.)</p>
+  <p><strong>If your request is urgent</strong></p>
+  <p>Submit this form and call:</p>
+  <ul>
+    <li><a href="tel:+447810750834">07810 750 834</a> between 9am and 6pm</li>
+    <li><a href="tel:+447827992603">07827 992 603</a> at other times</li>
```
ChrisBAshton, line 188: Looks like you're missing "It's urgent only if the public or the government faces immediate and significant financial, legal or physical risk".

app/controllers/content_change_requests_controller.rb
```diff
@@ -16,7 +16,7 @@ def parse_request_from_params
   def content_change_request_params
     params.require(:support_requests_content_change_request).permit(
       :title,
-      :reason,
+      :reason_for_request,
```
ChrisBAshton, line 19: I'd fix this up to the "add request reason field" commit.

ChrisBAshton (commented):
This looks like a good change 👍 I've left some comments below, mostly around commit management.

RE:
> This PR adds the requested changes up to "Time constraints".
> Time constraints are shared across multiple forms, and the Platform team is better placed to manage this update.

Is that true? I can only see the 'time_constraint' partial being used in one place, and that's [in content_change_requests](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/content_change_requests/_request_details.html.erb#L55), so I think you're safe to go ahead and change that...? (Other forms that use time constraints seem to [hardcode their fields](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/live_campaign_requests/_request_details.html.erb#L47-L54)).

And RE:

> There are also additional fields that need to be mapped on Zendesk.

Can you explain more about that step? E.g. link to a previous PR that does some mapping.
(Sorry - Platform Reliability technically own Support but we know no more about it than other teams, so any context you have is going to be useful!)

DilwoarH:
> This looks like a good change 👍 I've left some comments below, mostly around commit management.
> 
> 
> 
> RE:
> 
> > This PR adds the requested changes up to "Time constraints".
> 
> > Time constraints are shared across multiple forms, and the Platform team is better placed to manage this update.
> 
> 
> 
> Is that true? I can only see the 'time_constraint' partial being used in one place, and that's [in content_change_requests](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/content_change_requests/_request_details.html.erb#L55), so I think you're safe to go ahead and change that...? (Other forms that use time constraints seem to [hardcode their fields](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/live_campaign_requests/_request_details.html.erb#L47-L54)).
> 
> 
> 
> And RE:
> 
> 
> 
> > There are also additional fields that need to be mapped on Zendesk.
> 
> 
> 
> Can you explain more about that step? E.g. link to a previous PR that does some mapping.
> 
> (Sorry - Platform Reliability technically own Support but we know no more about it than other teams, so any context you have is going to be useful!)

This is the first time I've touched this repo so I know as much as you do.

I will try and detail the changes needed

app/views/content_change_requests/_request_details.html.erb
DilwoarH: Done

spec/features/content_change_requests_spec.rb
DilwoarH: done

app/views/content_change_requests/_request_details.html.erb
ChrisBAshton: This is still a thing @DilwoarH ^

DilwoarH:
@deborahchua @ChrisBAshton 

I have made all the changes to the forms. The only thing remaining to do is the mapping on Zendesk.

This is the first time I have touched this app, so I don't know enough to properly investigate how to get this done or where to start. I will therefore need to hand this off due to other priorities.

I will set up a call between the content team and us to ensure that we are all on the same page.

Details of the mapping needed on Zendesk.
```
- Custom “title” field
- Custom “reason for change” field
- Custom “subject area” field
- Custom “needs Welsh” field, based on “Welsh URL” field being filled
- “What’s the new or different information”, in Zendesk body
- “Why is change needed”, in Zendesk body
- Custom “Deadline” field (this may already happen)
- Custom “Not before” field (this may already happen)
- “Reason for deadline”, in Zendesk body
- “Who needs a copy of this request”, put into Zendesk cc field (this already happens)
```

Details of change request are added to the following doc:
https://docs.google.com/document/d/1s0OoBexbHXnDIx554EyudsdAYpAKI7C1p4VFBvqkW9Q/edit

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -38,7 +38,7 @@
         "Uprating - policy or money changes at the new tax year (6 April)",
         "Other"
       ]),
-      { include_blank: true } ,
```
ChrisBAshton: Please fix this up to the `add request reason field` commit.

spec/features/content_change_requests_spec.rb
```diff
@@ -83,6 +87,7 @@ def user_makes_a_content_change_request(details)
 
     fill_in "Title of request", with: details[:title] unless details[:title].nil?
     select details[:reason_for_change], from: "What’s the reason for the request?" unless details[:reason_for_change].nil?
+    select details[:subject_area], from: "What’s the subject area?" unless details[:subject_area].nil?
```
ChrisBAshton, line 94: Both of these `select`'s `from:` values [can refer to the ID/name of the select](https://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Actions#select-instance_method), rather than the text (makes it easier to rename the label of the field). So I think this would work...?

```suggestion
    select details[:reason_for_change], from: "reason_for_change" unless details[:reason_for_change].nil?
    select details[:subject_area], from: "subject_area" unless details[:subject_area].nil?
```

app/views/support/_time_constraint.html.erb
```diff
@@ -15,7 +15,7 @@
 
     <div class="form-group">
       <span class="form-label">
-        <%= r.label :not_before_date, "MUST NOT be published BEFORE" %>
+        <%= r.label :not_before_date, "Must not be published before" %>
```
ChrisBAshton, line 18: I can see we still need a new 'time' field for each of the above:

```
00:00
[TIME FIELD]
```

I think you're right that these changes are more involved - they'd need adding to the [`time_constraint_attributes` hash](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/controllers/content_change_requests_controller.rb#L23) which ultimately [populates the `WithTimeConstraint` module](https://github.com/alphagov/support/blob/b7e64c64a8e74b7221644d93bbed629ccf93b748/app/models/support/requests/with_time_constraint.rb#L13-L15). This uses the [validates_timeliness gem](https://github.com/adzap/validates_timeliness), which does appear to support times as well as dates.

Whilst these attributes do appear on the surface to also be used in [content_advice_requests_controller](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/controllers/content_advice_requests_controller.rb#L23) and [remove_user_requests_controller](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/controllers/remove_user_requests_controller.rb#L22), these are actually defined in their own templates ([1](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/content_advice_requests/_request_details.html.erb#L59-L75), [2](https://github.com/alphagov/support/blob/41cec714da8967acb16a41ac93bfea4fa3368852/app/views/remove_user_requests/_request_details.html.erb#L33-L42) respectively), so we should be free to add additional fields here and not worry about any knock-on effects to other support forms.

TLDR: to add 'time' fields, you'd need to:

* add a label & input for a 'time' field next to each of the 'date' fields above
* [update the `time_constraint_attributes`](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/controllers/content_change_requests_controller.rb#L23)
* add some validation logic for the new fields, into [TimeConstraint](https://github.com/alphagov/support/blob/7d344be1a0bce19d830adb8e399d7b97fd7a4fcd/app/models/support/requests/time_constraint.rb) (including [corresponding tests](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/spec/models/support/requests/time_constraint_spec.rb)) - I guess as a minimum you would just validate that the time is valid ("00:00" vs "banana"), but taking it another level you could check if either of the date fields are 'today' and then check if the time field is in the future or the past. (Otherwise the time field is largely irrelevant if the date is in the future, so long as the time is valid)

app/views/support/_time_constraint.html.erb
```diff
@@ -24,10 +24,15 @@
...
+        <%= r.label :time_constraint_reason, "Reason for deadline" %>
       </span>
+      <div class="help-block hint-block--md-6" id="reason-for-deadline-hint">
+        <p>For example, policy or campaign launch, start of the new tax year.</p>
+      </div>
       <span class="form-wrapper">
-        <%= r.text_area :time_constraint_reason, required: false, class: "input-md-6 form-control", rows: 6, cols: 50 %>
+        <%= r.text_area :time_constraint_reason, required: false, class: "input-md-6 form-control", rows: 6, cols: 50, aria: {
+          describedby: "reason-for-deadline-hint"
+        } %>
```
ChrisBAshton, line 35: Finally, it looks like you still need a 'legal deadline' field:

```
Is this a legal deadline?
[RADIO BUTTON FIELD]
```

This can be added in the same way as the other fields you've already added, as it doesn't factor into the 'time_constraint_attributes' at all.

ChrisBAshton (commented):
Cool, I've done a bit more research now into how this all works and there shouldn't be any additional 'Zendesk mapping' needed.

The `ContentChangeRequestTicket` model [inherits from `Zendesk::ZendeskTicket`](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/models/zendesk/ticket/content_change_request_ticket.rb#L3), which has a [`comment` method that uses this `comment_snippets` value](https://github.com/alphagov/support/blob/98f88ef5dba31485aae195939573ec8114d68866/app/models/zendesk/zendesk_ticket.rb#L16-L18) (it appears to just [join each comment as a string, with newlines](https://github.com/alphagov/support/blob/86259b99aea7625cf1f3a455011081fd0fd26df0/app/models/zendesk/snippet_collection.rb#L9).

It's this comment which is ultimately [sent to Zendesk](https://github.com/alphagov/support/blob/ad1967fa0f7b5cb7002181d5c43586711b64672c/app/models/zendesk/zendesk_tickets.rb#L10) via the [ZendeskTicketWorker](https://github.com/alphagov/support/blob/7d344be1a0bce19d830adb8e399d7b97fd7a4fcd/app/workers/zendesk_ticket_worker.rb), which is [created in the first place by the `RequestsController`](https://github.com/alphagov/support/blob/c1f1430c2ff93e8461b88e1560699fb9d95979b5/app/controllers/requests_controller.rb#L36) that the [`ContentChangeRequestsController` inherits from](https://github.com/alphagov/support/blob/7012da5f0f798a357b6429b9eaeb6544abb77927/app/controllers/content_change_requests_controller.rb#L1).

For the outstanding fields that still need adding, I've put some suggestions below. But I agree it's probably fine to add the 'time' fields in a follow-up PR, which Platform Reliability could potentially help with if needed.

spec/features/content_change_requests_spec.rb
DilwoarH: This doesn't work unfortunately

ChrisBAshton:
NB I noticed an existing issue (as a result of reviewing your PR), which I've fixed in #1092. You'll need to rebase your branch I'm afraid!

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -55,7 +55,13 @@
 <%= render partial: "support/time_constraint", locals: { f: f } %>
 
 <div class="alert alert-info alert-block">
-  <p>If your request is urgent, please fill in this form and call <%= @primary_contact_details[:urgent_department_content_change][:phone] %> during office hours, or <%= @primary_contact_details[:urgent_mainstream_content_change][:phone] %> out of hours. (It's ONLY urgent when the public or the government is facing immediate and significant financial, legal or physical risk.)</p>
+  <p><strong>If your request is urgent</strong></p>
+  <p>Submit this form and call:</p>
+  <ul>
+    <li><a href="tel:+447810750834"><%= @primary_contact_details[:urgent_department_content_change][:phone] %></a> between 9am and 6pm</li>
+    <li><a href="tel:+447827992603"><%= @primary_contact_details[:urgent_mainstream_content_change][:phone] %></a> at other times</li>
```
ChrisBAshton, line 188: We want to keep these numbers out of `main`. I suggest dropping the link markup entirely, otherwise you'll need to do something like `<%= foo.sub("0","+44").delete(" ") %>

ChrisBAshton (approved):
LGTM 👍  We recognise that there are still two features to add ([1](https://github.com/alphagov/support/pull/1087#discussion_r1097273049) and [2](https://github.com/alphagov/support/pull/1087#discussion_r1097273282)) but these can be added in future PRs.

DilwoarH:
Waiting on Content Team to approve before merge

ChrisBAshton (commented):
Good start, @MahmudH 👍 please see comments below, and also update the Trello card URL and screenshots in this PR description. Thanks!

app/views/support/_time_constraint.html.erb
```diff
@@ -22,6 +22,24 @@
...
     </div>
 
+    <div class="form-group">
+      <span class="form-label">
+        <%= f.label :is_legal_deadline do %>
+          Is this a legal deadline?
+        <% end %>
+      </span>
+      <span class="form-wrapper">
+        <%= f.collection_radio_buttons :is_legal_deadline, f.object.is_legal_deadline_options, :last, :last, required: false, aria: { required: false } do |builder| %>
```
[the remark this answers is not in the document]
ChrisBAshton: The content people have now [confirmed this should be required](https://trello.com/c/aPOXVkPS/3117-complete-changes-to-content-requests-form-in-support-app-3#comment-6423ffb7dfc420d890c6d868).

app/views/support/_time_constraint.html.erb
```diff
@@ -15,7 +15,34 @@
 
     <div class="form-group">
       <span class="form-label">
-        <%= r.label :not_before_date, "MUST NOT be published BEFORE" %>
+        <%= r.label :needed_by_time, "Time" %>
```
ChrisBAshton, line 18: The labels should [now be changed](https://trello.com/c/aPOXVkPS/3117-complete-changes-to-content-requests-form-in-support-app-3#comment-6423ffb7dfc420d890c6d868) to:

- Deadline
- Time this must be published by
- Must not be published before
- Time this must not be published before

spec/features/content_change_requests_spec.rb
```diff
@@ -58,7 +58,9 @@
       url: "http://gov.uk/X",
       related_urls: "http://gov.uk/welsh",
       needed_by_date: "31-12-#{next_year}",
+      needed_by_time: "13:00",
       not_before_date: "01-12-#{next_year}",
+      not_before_time: "18:00",
```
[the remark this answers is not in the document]
ChrisBAshton: @MahmudH did you see this comment? And the one below.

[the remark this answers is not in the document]
ChrisBAshton: Looks like this is done now: https://github.com/alphagov/support/blob/472-changes-to-content-requests-form-in-publisher/app/controllers/content_change_requests_controller.rb

spec/models/support/requests/time_constraint_spec.rb
```diff
@@ -32,6 +36,57 @@ def as_str(date)
         expect(constraint).to have_at_least(1).error_on(:not_before_date)
       end
 
+      it "doesn't allow the 'not before' time to be set after the 'needed by' time, if the 'not before' date and 'not before' time are the same" do
```
ChrisBAshton, line 39: Loving these tests! ⭐

ChrisBAshton (commented):
Hi Mahmud - it's shaping up well. I have one blocking comment below (split into three suggestions), the rest are just nitpicks 👍

app/views/support/_time_constraint.html.erb
```diff
@@ -6,13 +6,40 @@
...
+    </div>
+
+    <div class="form-group">
+      <span class="form-label">
+        <%= f.label :is_legal_deadline do %>
+          Is this a legal deadline?
+        <% end %>
+      </span>
+      <span class="form-wrapper">
+        <%= f.collection_radio_buttons :is_legal_deadline, f.object.is_legal_deadline_options, :last, :last, required: true, aria: { required: false } do |builder| %>
```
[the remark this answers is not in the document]
MahmudH: The collection radio buttons don't take an argument of `required` so I've removed it.

spec/support/app_actions.rb
```diff
@@ -77,8 +77,11 @@ def doctype_summary_results
   end
 
   def user_fills_out_time_constraints(details)
-    fill_in "Deadline (if you have one)", with: details[:needed_by_date]
+    fill_in "Deadline", with: details[:needed_by_date]
+    fill_in "Time this must be published by", with: details[:needed_by_time]
     fill_in "Must not be published before", with: details[:not_before_date]
+    choose "No"
```
ChrisBAshton, line 83: This should be removed from the `Add time input fields and update date labels` commit (I think it refers to the "Is this a legal deadline?" question, which hasn't been added yet.

app/views/content_change_requests/_request_details.html.erb
```diff
@@ -81,36 +88,37 @@
...
-      Are any Welsh URLs affected? <abbr title="required">*</abbr>
-    <% end %>
+    <%= f.label :related_urls, "Are any Welsh URLs affected?" %>
   </span>
   <p class="help-block hint-block--md-6" id="related-urls-hint">
     Include any Welsh translations. List URLs, one per line.
   </p>
   <span class="form-wrapper">
-    <%= f.text_area :related_urls, required: true, aria: { required: true }, class: "input-md-6 form-control", rows: 4, cols: 50, aria: {
+    <%= f.text_area :related_urls, aria: { required: true }, class: "input-md-6 form-control", rows: 4, cols: 50, aria: {
```
ChrisBAshton, line 121: If this isn't required, we should remove the aria:

```suggestion
    <%= f.text_area :related_urls, class: "input-md-6 form-control", rows: 4, cols: 50, aria: {
```

Also the bit we've just removed was probably conflicting with the `aria: { describedby: "related-urls-hint"` bit (i.e. you have two values for `aria`, I'm not sure how Rails deals with that).

app/views/support/_time_constraint.html.erb
```diff
@@ -22,6 +22,24 @@
...
     </div>
 
+    <div class="form-group">
+      <span class="form-label">
+        <%= f.label :is_legal_deadline do %>
+          Is this a legal deadline?
+        <% end %>
+      </span>
+      <span class="form-wrapper">
+        <%= f.collection_radio_buttons :is_legal_deadline, f.object.is_legal_deadline_options, :last, :last, required: false do |builder| %>
```
ChrisBAshton: Is `required: false` necessary, or should this just be removed entirely? I imagine `false` is the default so doesn't need specifying?

EDIT: you remove it in `Validate times and radio button` 👍 So just needs removing from here too (the "Add radio button" commit)

app/models/support/requests/content_change_request.rb
```diff
@@ -22,6 +22,13 @@ def self.label
       def self.description
         "Request changes to GOV.UK content managed by GDS content designers"
       end
+
+      def is_legal_deadline_options
```
ChrisBAshton, line 26: Nitpick: I'd rename the commit to `Add "Is this a legal deadline?" question`, as "Add radio button" is a bit vague.

spec/models/support/requests/time_constraint_spec.rb
```diff
@@ -43,6 +98,22 @@ def as_str(date)
...
+        %i[needed_by_time not_before_time].each do |attribute|
+          should allow_value("14:00").for(attribute)
+          should_not allow_value("25:00").for(attribute)
+          should_not allow_value("14:60").for(attribute)
+          should_not allow_value("14").for(attribute)
+          should_not allow_value("1400").for(attribute)
+          should_not allow_value("14:00 PM").for(attribute)
+          should_not allow_value("2 : 00").for(attribute)
+          # These values should never happen in practice because the dropdown menu uses 24 hour time,
+          # but they're still valid
```
ChrisBAshton, line 112: Thanks for adding these 👍

app/models/support/requests/time_constraint.rb
```diff
@@ -4,11 +4,27 @@ module Support
...
+      attr_accessor :not_before_date, :not_before_time, :needed_by_date, :needed_by_time, :time_constraint_reason, :is_legal_deadline
+
+      validates :is_legal_deadline, inclusion: %w[yes no], allow_blank: true
 
       validates_date :needed_by_date, allow_nil: true, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, allow_nil: true, allow_blank: true, on_or_after: :today
       validates_date :not_before_date, on_or_before: :needed_by_date, unless: proc { |c| c.needed_by_date.blank? || c.not_before_date.blank? }
+
+      validates_time :not_before_time, allow_nil: true, allow_blank: true
+      validates_time :needed_by_time, allow_nil: true, allow_blank: true
```
ChrisBAshton, line 16: What happens if the user doesn't fill in the field? I imagine we only need to `allow_nil` or `allow_blank`, not both (probably the latter?).

ChrisBAshton (commented):
This is really shaping up, @MahmudH - only some minor comments below.

ChrisBAshton (approved):
Awesome work ⭐ This is ready to go!

Nitpick: last commit message (`- The validation is not needed here as there will be not nil passed through`) doesn't read very clearly: perhaps `This validation is not needed as nil is never passed. If left unfilled, value will be a blank empty string` instead.

app/controllers/content_change_requests_controller.rb
```diff
@@ -16,11 +16,14 @@ def parse_request_from_params
...
       :title,
+      :reason_for_change,
+      :subject_area,
       :details_of_change,
+      :why_is_change_needed,
       :url,
       :related_urls,
       requester_attributes: %i[email name collaborator_emails],
-      time_constraint_attributes: %i[not_before_date needed_by_date time_constraint_reason],
+      time_constraint_attributes: %i[not_before_date needed_by_date time_constraint_reason needed_by_time not_before_time],
```
ChrisBAshton, line 26: Hey 👋 apparently these two new fields aren't making it through to Zendesk. I think we'll need to add them to [zendesk_ticket.rb](https://github.com/alphagov/support/blob/98f88ef5dba31485aae195939573ec8114d68866/app/models/zendesk/zendesk_ticket.rb#L48-L58).

app/models/zendesk/zendesk_ticket.rb
```diff
@@ -50,6 +50,8 @@ def base_comment_snippets
         [
           Zendesk::LabelledSnippet.new(on: self, field: :needed_by_date, label: "Needed by date"),
           Zendesk::LabelledSnippet.new(on: self, field: :not_before_date, label: "Not before date"),
+          Zendesk::LabelledSnippet.new(on: self, field: :needed_by_time, label: "Needed by time"),
+          Zendesk::LabelledSnippet.new(on: self, field: :not_before_time, label: "Not before time"),
```
ChrisBAshton, line 54: You may need to define two new methods in addition to this- see https://github.com/alphagov/support/blob/98f88ef5dba31485aae195939573ec8114d68866/app/models/zendesk/zendesk_ticket.rb#L20-L30

PR #1103, opened 2023-03-23
Post-replatforming cleanup.

sengi:
- Remove Jenkinsfile and Procfile; we no longer build/test on Jenkins and we no longer run on Foreman (or Heroku, or anything else that uses Procfile).
- Update the README to remove references to govuk-puppet etc.
- Fix the rake task for running the linters.
- Remove the Unicorn config so that we're ready for when Unicorn is removed in a future govuk_app_config release.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,8 +3,6 @@
 .github
 .gitignore
 Dockerfile
-Jenkinsfile
-Procfile
 README.md
 coverage
 docs
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,0 @@
-#!/usr/bin/env groovy
-
-library("govuk")
-
-node {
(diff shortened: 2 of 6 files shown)


ChrisBAshton:
Hi @sengi are Replatforming planning to roll this out across all repos or is it up to individual teams?

sengi:
Good question. This PR is mainly just me figuring out what common cleanup might be needed, which is why it's doing several (related but distinct) things that could be done atomically.

I'm leaning towards doing as much as we can in bulk from Replatforming since it'll save a lot of effort and confusion for dev teams, but we need to do a final sweep of Jenkins rake task jobs before we can get rid of all the deployments to Jenkins.

Apologies for the confusion around this and thanks for bearing with us while we figure out the least-hassle way forward.

PR #1123, opened 2023-05-24
Fix url validation

MuriloDalRi:
Brings change made in https://github.com/alphagov/support/pull/1111 to `main` and add a test for it.


AgaDufrat (approved):
LGTM. 
I wonder if we should have followed [A change from an external contributor docs](https://docs.publishing.service.gov.uk/manual/merge-pr.html#a-change-from-an-external-contributor) to trigger the CI run on the original PR.

PR #1170, opened 2023-08-24
Add functionality to set Zendesk custom fields and form template

AgaDufrat:
## Context

The current Content Change Request form also doesn't make the most of Zendesk. Making better use of custom fields would allow content design teams to improve internal reporting (e.g. improvement work tickets vs factual inaccuracy tickets).
The form template and custom fields were [created in the UI](https://support.zendesk.com/hc/en-us/articles/4408838961562).

[Trello card](https://trello.com/c/9j3DslWM/3176-apply-zendesk-mapping-to-the-content-requests-form-5)

## What

- Adds ability to set custom fields and form template in Zendesk
- Sets the following custom fields for Content change request tickets:
    - [CR] Reason for the request (required)
    - [CR] Subject Area(required)
    - [CR] Deadline
    - [CR] Do not publish before
    - [CR] Deadline (time)
    - [CR] Do not publish before (time)
- Sets 'Content Request' form template for Content change request tickets

| Before | After |
| --- | --- |
| Screenshot 2023-08-24 at 10 23 52 | Screenshot 2023-08-23 at 15 16 11 |

Ticket examples:

- [Ticket without time constraint](https://govuk.zendesk.com/agent/tickets/5460403)
- [Ticket with time constraint](https://govuk.zendesk.com/agent/tickets/5460411)

## Review notes
- Best reviewed commit by commit (see message for more details)
- Since the custom fields information needs to be hardcoded, I decided to use a YAML file to store it. Loading ruby classes from a YAML file introduces a level of abstraction  but has the following benefits:
    - allows us to store all information about the Zendesk Custom fields in a YAML file (the app doesn't seem to have access to the custom fields end point via Zendesk API)
    - specific Ticket classes don't need to know about field type
    - CustomField class is not responsible for validating/formatting values of different types
    - makes it easy to extend should we want to populate other type of fields such as multi-select.

## How to test it
- deploy the branch to integration
- visit https://support.integration.publishing.service.gov.uk/content_change_request/new
- try different options
- find submitted ticket in [Zendesk](https://govuk.zendesk.com/agent/search/1) 
- try other forms

---


app/models/support/requests/content_change_request.rb
```diff
@@ -22,6 +22,50 @@ def self.label
       def self.description
         "Request changes to GOV.UK content managed by GDS content designers"
       end
+
+      def self.reason_for_change_options
+        [
+          "Factual inaccuracy",
```
robinjam, line 28: Is there any reason why these are still hardcoded here instead of being read from the new YAML file?

robinjam: (Sorry, meant to start a review rather than posting a single comment! 🤦‍♂️)

spec/fixtures/zendesk_custom_fields_data.yml
```diff
@@ -0,0 +1,12 @@
...
+  name: "Deadline"
+  id: 123
+- !ruby/object:Zendesk::CustomFieldType::DropDown
+  name: "Reason for the request"
+  id: 456
+  options:
+    name_tag1: Factual inaccuracy
+- !ruby/object:Zendesk::CustomFieldType::Text
+  name: "URLs"
+  id: 789
```
robinjam, line 12: No newline at end of file (ignore me if there's a reason for this!)

spec/models/zendesk/custom_field_type/date_field_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(subject.prepare_value(input)).to eq("2023-08-02")
+    end
+
+    it "retruns date in the ISO 8601 date format (extended) given ISO 8601 formatted date with time and timezone" do
+      input = "2023-08-02T20:49:15Z"
+
+      expect(subject.prepare_value(input)).to eq("2023-08-02")
+    end
+
+    it "retruns date in the ISO 8601 date format (extended) given ISO 8601 formatted date string with time" do
```
robinjam, line 25:
```suggestion
    it "returns date in the ISO 8601 date format (extended) given ISO 8601 formatted date string with time" do
```

spec/models/zendesk/custom_field_type/date_field_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      input = "2023-08-02T18:21:06"
+
+      expect(subject.prepare_value(input)).to eq("2023-08-02")
+    end
+
+    it "retruns date in the ISO 8601 date format (extended) given RFC 2822 formatted string" do
```
robinjam, line 31:
```suggestion
    it "returns date in the ISO 8601 date format (extended) given RFC 2822 formatted string" do
```

robinjam (commented):
This is amazing Aga 🌟 

Couple of nitpicks (plus the question above), but it's looking great

app/models/support/requests/content_change_request.rb
AgaDufrat: That's a great idea, thanks! ⭐  
It means I can delete the config test that prevents them going out of sync.

robinjam (approved):
LGTM, nice work!

PR #1175, opened 2023-09-01
Fix bug when filtering anonymous feedback

MuriloDalRi:
The current code won't show the feedback for an organisation unless a '/' URL is specified.

https://trello.com/c/b8wn2ZQK/3252-improve-anonymousfeedback-search-behaviour-in-support-app-2


ChrisBAshton (commented):
Thanks @MuriloDalRi - this fix looks nice and simple, but it's missing two things:

1. A test.
2. Removal of whatever code is causing the current behaviour of a random URL being added to the 'URL' field (like this: "/e38e5fb202880093ee51").

MuriloDalRi:
> Thanks @MuriloDalRi - this fix looks nice and simple, but it's missing two things:
> 
> 1. A test.
> 2. Removal of whatever code is causing the current behaviour of a random URL being added to the 'URL' field (like this: "/e38e5fb202880093ee51").

@ChrisBAshton I've added a test now. 
Do you mean a random URL being added to the form or to the browser's address bar? If you filter with an empty URL field it will now just add a /

ChrisBAshton:
Thanks, the test looks good. Yep, we should remove the associated code as it now will no longer ever get evaluated. Aga reckons it's the `SecureRandom.hex(10)` here that needs removing:

https://github.com/alphagov/support/blob/main/app/models/support/requests/anonymous/paths.rb#L7

ChrisBAshton (approved):
I've just tried it out locally - and yes, looks like it will be difficult to unpick the SecureRandom. I believe it's used in the [saved paths feature](https://github.com/alphagov/support/blob/27ab487a252c7cc68b209eeaf3695cd2383e2c30/app/controllers/anonymous_feedback_controller.rb#L33-L43). By removing the hex, we're no longer storing at a unique key in Redis, and locally I was finding the search of `/` was "sticking" and making it impossible to try to submit an empty form, which would kind of solve this issue but could also be hiding a more gnarly bug under the surface.

Your fix seems to work in a nice low-risk way, so happy to approve 👍

MuriloDalRi:
> I've just tried it out locally - and yes, looks like it will be difficult to unpick the SecureRandom. I believe it's used in the [saved paths feature](https://github.com/alphagov/support/blob/27ab487a252c7cc68b209eeaf3695cd2383e2c30/app/controllers/anonymous_feedback_controller.rb#L33-L43). By removing the hex, we're no longer storing at a unique key in Redis, and locally I was finding the search of `/` was "sticking" and making it impossible to try to submit an empty form, which would kind of solve this issue but could also be hiding a more gnarly bug under the surface.
> 
> Your fix seems to work in a nice low-risk way, so happy to approve 👍

Thanks @ChrisBAshton

PR #1183, opened 2023-09-25
Fix release version in footer

deborahchua:
This has been broken since we moved to EKS and the SHA is no longer provided.

We can use the [SENTRY_RELEASE](https://github.com/alphagov/govuk-helm-charts/blob/077117e7f40c096a958f9e0b853a2bd76cf43f26/charts/generic-govuk-app/templates/deployment.yaml#L92C1-L92C1) environment variable which is defined for every app and takes its value from the container image's tag.

If for whatever reason the image tag is not defined it will display `null`, which should make it clearer (instead of displaying `development`) that this data is not being pulled in successfully.

Before:
Screenshot 2023-09-25 at 14 37 33

After (tested in Integration):
Screenshot 2023-09-25 at 14 38 09

Trello card: https://trello.com/c/uoKDUcbV/3182-fix-broken-version-footer-in-publishing-apps-2

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -56,6 +56,6 @@
   <%= yield %>
 <% end %>
 
-<% content_for(:footer_version) { CURRENT_RELEASE_SHA } %>
+<% content_for :footer_version, ENV.fetch("SENTRY_RELEASE", "null")[0..18] %>
 
 <%= render template: 'layouts/govuk_admin_template' %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -56,6 +56,6 @@
   <%= yield %>
 <% end %>
 
-<% content_for(:footer_version) { CURRENT_RELEASE_SHA } %>
+<% content_for :footer_version, ENV.fetch("SENTRY_RELEASE", "null") %>
```
AgaDufrat, line 59: What do you think of only showing a short SHA (first 10 characters) like it was before?

```suggestion
<% content_for :footer_version, ENV.fetch("SENTRY_RELEASE", "null")[0..18] %>
```

deborahchua:
> Looks good and nice idea of using the SENTRY_RELEASE! 

I cannot take the credit - this was @robinjam 's suggestion! 🥇

PR #1202, opened 2023-10-13
Improve email check regex

MuriloDalRi:
Use a more robust regex to check emails.

https://trello.com/c/oFWuEIyv/3318-improve-email-validation-in-support-app-3

### DIFF
--- a/app/models/support/requests/requester.rb
+++ b/app/models/support/requests/requester.rb
@@ -7,9 +7,11 @@ module Support
       attr_accessor :name
       attr_reader :email
 
+      VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
+
       validates :email, presence: true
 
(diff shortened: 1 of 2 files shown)


AgaDufrat (commented):
Can we add some test (at lease for the example in the Sentry error)?

MuriloDalRi:
> Can we add some test (at lease for the example in the Sentry error)?

Done

AgaDufrat (approved):

app/models/support/requests/requester.rb
```diff
@@ -7,6 +7,8 @@ class Requester
       attr_accessor :name
       attr_reader :email
 
+      VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
+
       validates :email, presence: true
 
       validates :email, format: { with: /@/ }
```
AgaDufrat, line 14: Do we also need to change it here?

PR #1205, opened 2023-10-18
Update dependency to solve alerts

MuriloDalRi:
This gem has been pretty much abandoned and we need to do the work to ultimately replace it with a component available in the design system, however this is a quick fix for the alerts we've been getting.


marcpomfret (approved):
ok

PR #1207, opened 2023-10-19
Add fields to the 'Request a new campaign' form

MuriloDalRi:
Add HMG code and strategic planning code fields to the request a new campaign form.

https://trello.com/c/NoFOCDwj/3320-add-fields-to-the-request-a-new-campaign-form-in-support-app-2


AgaDufrat (approved):
Looks good. I noticed that in the request there was a space on either side of the colon ` : `, which I think is rather usual formatting. Given that we might be moving to Design System 🤞 I think it's good to go, thanks! 🚀

PR #1248, opened 2024-01-17
Fix validation-related bugs for organisation field in "Create new user" request form

floehopper:
This fixes a couple of bugs:
* A `NoMethodError` was being raised for any server-side validation error for this form
* The selected organisation was not being retained after a server-side validation error

Luckily both of these were masked from most users by the client-side validation and this may explain why we haven't seen any `NoMethodError` exceptions in the wild.

### DIFF
--- a/app/views/create_new_user_requests/_requested_user_details.html.erb
+++ b/app/views/create_new_user_requests/_requested_user_details.html.erb
@@ -28,7 +28,7 @@
       <% end %>
     </span>
     <span class="form-wrapper">
-      <%= r.select :organisation, options_for_select(organisation_options), { include_blank: true }, multiple: false, class: "select2 form-control", type: "organisation" %>
+      <%= r.select :organisation, options_for_select(organisation_options, r.object.organisation), { include_blank: true }, multiple: false, class: "select2 form-control", type: "organisation" %>
     </span>
   </div>
 <% end %>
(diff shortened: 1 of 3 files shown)


MuriloDalRi (approved):

app/views/create_new_user_requests/_requested_user_details.html.erb
```diff
@@ -28,7 +28,7 @@
       <% end %>
     </span>
     <span class="form-wrapper">
-      <%= r.select :organisation, options_for_select(@organisations_list), { include_blank: true }, multiple: false, class: "select2 form-control", type: "organisation" %>
```
AgaDufrat, line 31: Could you please mention/justify the removal of `select2` class, which I believe will remove the option to search for a department in the list.

AgaDufrat (commented):
Good spot. Controller helper method seems like a good approach. I left a comment around the removal of `select2` class.

app/views/create_new_user_requests/_requested_user_details.html.erb
floehopper: Oops! Thanks for spotting it - I removed it while debugging and forgot to re-add it! 😬

floehopper: Addressed in https://github.com/alphagov/support/pull/1248/commits/0cd495b655483865b0abca4631caed58b7f5c466

AgaDufrat (approved):

PR #1252, opened 2024-01-18
Move "Create new user" and "Change existing user" request forms to use GOV.UK Design System

floehopper:
Trello: https://trello.com/c/aQiPdsiP

PR #1253, opened 2024-01-22
Move from libsass -> dart-sass

floehopper:
Since LibSass is now [deprecated](https://sass-lang.com/blog/libsass-is-deprecated/), this is a change that needs making anyway. However, my actual motivation is that ~I'm hoping it will fix~ it fixes a `SassC::SyntaxError` that I'm seeing after adding `govuk_publishing_components` when running `RAILS_ENV=production rails assets:precompile` in order to start moving the app to use the GOV.UK Design System. See this [build failure](https://github.com/alphagov/support/actions/runs/7574079838/job/20627675163) which occurred against the code in [this draft PR](https://github.com/alphagov/support/pull/1252).

I followed [these instructions][1] in the GOV.UK Developer Docs to make the change. Once it's merged, there's also a corresponding change to [govuk-docker](https://github.com/alphagov/govuk-docker) that will need to me made - see [this PR](https://github.com/alphagov/govuk-docker/pull/714).

I've tested this locally using the modified version of `govuk-docker` and verified that I can make changes to CSS and see them reflected in the rendered pages. I've also run `RAILS_ENV=production rails assets:precompile` locally and I don't see any errors or warnings. However, note that the instructions told me to suppress warnings in dependencies and before I did that, I did see some deprecation warnings before I did that which doesn't seem ideal.

[1]: https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html


procfile.dev
chrisroos, line 1: I think this should be Procfile.dev (to match the filename in bin/dev).

procfile.dev
```diff
@@ -0,0 +1,2 @@
+web: bin/rails server -p 3070
```
chrisroos, line 1: I think this should be port 3000, as per the comment in [the docs](https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html#watch-mode):

> Note, the port number should match the one previously specified, in startup.sh and or docker-compose.yml, for running your Rails server.

bin/dev
chrisroos, line 1: Does this work for you in Docker? Although `gem list foreman` says that I have Foreman 0.8.7 installed, the binary isn't in the `$PATH` so `exec foreman ...` doesn't actually do anything.

Gemfile.lock
chrisroos, line 1: Did you run `bundle install` on your machine or in Docker? If I make the same changes to the `Gemfile` and run `bundle install` in Docker then my `Gemfile.lock` contains some differences to yours (e.g. some Gems include a `x86_64-linux` suffix and in the `PLATFORMS` section `ruby` has been removed and `x86_64-linux` has been added).

procfile.dev
floehopper: Addressed in https://github.com/alphagov/support/pull/1253/commits/610e25a519a80a6db8e3918bd38c5ff2e1b7d3cc

procfile.dev
floehopper: Addressed in https://github.com/alphagov/support/pull/1253/commits/610e25a519a80a6db8e3918bd38c5ff2e1b7d3cc

Gemfile.lock
floehopper: We've addressed this in #1254 and by adding the following note to the commit in this PR:

> Note that running `bundle install` has removed the `ruby` platform. We think this is because `sass-embedded`, a dependency of `dartsass-rails`, doesn't have a non-platform-specific variant of the gem and so bundler can no longer support the `ruby` platform.

bin/dev
floehopper: It does, but I'm a bit uncertain about the state of my local Docker environment. We're going to continue to investigate before merging.

floehopper:
> However, my actual motivation is that I'm hoping it will fix a `SassC::SyntaxError` that I'm seeing after adding `govuk_publishing_components` when running `RAILS_ENV=production rails assets:precompile` in order to start moving the app to use the GOV.UK Design System. See this [build failure](https://github.com/alphagov/support/actions/runs/7574079838/job/20627675163) which occurred against the code in [this draft PR](https://github.com/alphagov/support/pull/1252).

I have now rebased the branch in [this draft PR](https://github.com/alphagov/support/pull/1252) against this branch and confirmed that it does indeed fix the `SassC::SyntaxError`. 🎉  I've updated this PR description accordingly.

bin/dev
chrisroos: I've just rebuilt my dev environment and `bin/dev` now works for me too :partying_face:

PR #1255, opened 2024-01-23
Add form for people to report problems with new GOV.UK search

chrisroos:
https://trello.com/c/odKZPy7o

GOV.UK are trialling a new search engine. This new form allows people to report problems they encounter with this new search engine.

### DIFF
--- a/app/controllers/report_an_issue_with_govuk_search_results_requests_controller.rb
+++ b/app/controllers/report_an_issue_with_govuk_search_results_requests_controller.rb
@@ -1,0 +1,24 @@
+class ReportAnIssueWithGovukSearchResultsRequestsController < RequestsController
+protected
+
+  def new_request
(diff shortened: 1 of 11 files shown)


floehopper (approved):
This all looks good to me! 👍 

One very minor thing is that I'm not sure the `#user_reports_issue_with_search_results` private method in the new feature spec is adding much since it's only called from one scenario. In my current branch, I've inlined such methods. However, I suspect you've just copied what was there already, so there's an argument that what you've done is more consistent. Anyway, I'll leave you to decide!

chrisroos:
Thanks @floehopper :+1: 

I'm going to leave the `#user_reports_issue_with_search_results` private method as that appears to follow the pattern we're using in the other feature specs. I agree that it's not particularly necessary given that it's only used in one example, though.

PR #1259, opened 2024-01-25
 Move "Create new user" request form to use GOV.UK Design System

floehopper:
Trello: https://trello.com/c/aQiPdsiP

This introduces a new layout based on the GOV.UK Design System [admin layout](https://components.publishing.service.gov.uk/component-guide/layout_for_admin) and makes use of it in the "Create new user" request form following the design created by Calum, our Designer. I've also used the [X-GOVUK Form Builder for Ruby on Rails](https://govuk-form-builder.netlify.app/) to generate the form elements. And I've used the [accessible-autocomplete component](https://github.com/alphagov/accessible-autocomplete) to progressively enhance the organisation drop-down to allow users to search for the relevant organisation. This should make it a lot easier to start moving other pages to use the GOV.UK Design System. See screenshots below.

* The navigation items in the header have changed slightly in the new layout:
  * A new link to the home page has been added. Although the logo in the old design does the same thing; we felt this was clearer.
  * A "Switch app" link which takes you to the Signon dashboard. This is provided in _some_ of the other GOV.UK Publishing apps and feels like a useful addition and a small step towards some consistency.
* I haven't included breadcrumb links for the new version of the page, because I've introduced a new link to the home page in the global navigation and the Design System recommends not adding breadcrumbs if a link to the parent page already exists.
* I've made some changes to the validation to make it more idiomatic so that it plays nicely with [the built-in functionality in the form builder](https://govuk-form-builder.netlify.app/introduction/error-handling/). Since the validation for the "Change existing user" request form was partially shared, I made similar changes to that form too even though we haven't yet moved that form to use the GOV.UK Design System. If anything, I think these changes are an improvement even in the old design.
* I've moved the validation error messages to the locale file, because they were quite verbose and would have cluttered up the model code considerably. When I did this I also made the specs more robust by not hard-coding the expected error messages.
* This PR also includes a few copy changes requested by our Content Designer, Etain.
* I'm pretty confident I've re-implemented all the important functionality from the old layout to the new layout. Some things I haven't yet addresses are the `content_for(`:sub_header`) and `content_for(:footer_version)`, but there isn't yet any need for the former and there doesn't seem to be any precedence for the latter in other apps yet.
* Due to the heavy use of inheritance in this app, I've ended up having to override a bunch of controller methods and view templates to make this work. Hopefully once all the pages have been converted (!), we'll be able to simplify things again.
* Minor: I think it might be nice to rename `content_for(:header)` to `content_for(:heading)`, because I think that's a more intention-revealing name, but I decided that was too pervasive a change to make as part of this PR.
* See individual commit notes for more details.

### Create new user request - blank form (desktop)
Screenshot 2024-01-25 at 16 15 06

### Create new user request - blank form (mobile)
Screenshot 2024-01-25 at 16 16 00

### Create new user request - yes/no radio button in conjunction with text area
Screenshot 2024-01-25 at 16 16 35

### Create new user request - searchable organisation dropdown
Screenshot 2024-01-25 at 16 17 28

### Create new user request - validation errors
Screenshot 2024-01-25 at 16 18 06

### Create new user request - validation error on additonal comments
Screenshot 2024-01-25 at 16 18 40

Supersedes #1252.


jlewr:
Might be an issue with prototypes/screenshots - but the font seems off on the dropdown? (Searchable organisation dropdown)

floehopper:
> Might be an issue with prototypes/screenshots - but the font seems off on the dropdown? (Searchable organisation dropdown)

Addressed in https://github.com/alphagov/support/pull/1259/commits/1fbfd66928d013620fed18455ae952e226fffc47

floehopper:
@chrisroos Etain, Calum & Joseph have all said they're happy with this from a UX, design & copy point-of-view. I fixed a couple of things up on Friday morning based on their feedback, but haven't updated the screenshots in the PR description.

chrisroos:
@jkempster34 pointed out that [he used the form builder in Transition but that it was suggested he replace it with govuk_publishing_components](https://github.com/alphagov/transition/pull/1392#issuecomment-1717918215). I'm following up in Slack to see whether it's OK for us to use the form builder in this app.

chrisroos:
@richardTowers has confirmed that we should avoid using the form builder for now so I'm going to strip it out of this PR. Richard isn't against using it forever but is keen that we have a wider discussion before deciding whether or not to adopt it for publishing apps.

chrisroos:
I've just discussed this with @jlewr and we've agreed to park this PR for now as we think the time would be better spent moving this form into Signon. We've also agreed to look at converting the homepage of Support into the design system so that we benefit from the work @floehopper has done in this PR.

PR #1264, opened 2024-01-30
Move homepage to use GOV.UK Design System

chrisroos:
Trello: https://trello.com/c/SvasRpas

This builds on the first part of @floehopper's work in #1259 in order to convert the homepage to use the Design System. This should pave the way for the rest of the Support app to be converted to the Design System in future.

See the individual commits for more detailed information.

# Screenshot before

before

# Screenshot after

after


ChrisBAshton (approved):
LGTM - gave it a try locally too ✅ 

One request: in your commit, you've said `The "Make new request" dropdown hasn't been ported to the navigation in the new Design System layout. As agreed with @jlewr.`. I'd like to know more about that. Could you elaborate in the commit message, about the reasons why it wasn't carried over (e.g. is it going to be brought in via a later PR?)

chrisroos:
Thanks @ChrisBAshton :+1: 

> One request: in your commit, you've said The "Make new request" dropdown hasn't been ported to the navigation in the new Design System layout. As agreed with @jlewr.. I'd like to know more about that. Could you elaborate in the commit message, about the reasons why it wasn't carried over (e.g. is it going to be brought in via a later PR?)

I've updated the commit message with the following in the hope that it explains the rationale for omitting the dropdown:

> As agreed with @jlewr the "Make new request" dropdown hasn't been ported to the navigation in the new Design System layout because we don't think it adds any value over the list of links that appear on the homepage. The dropdown remains in the existing layout so that users can still jump between forms directly rather than going via the homepage. We're hoping that reducing the number of forms in the Support app, and improving those that remain, will mean we won't have to implement the dropdown as we convert more pages to the Design System because the forms will do a better job of routing users to the right place.

Do you think that's clear enough?

floehopper:
If any pages redirect to the home page with messages set in the flash, those messages will no longer be displayed.

PR #1294, opened 2024-03-11
Create new user form improvements

jkempster34:
https://trello.com/c/2B5qTHlZ

image

### Update label for additional comments
We are going to add additional questions so that the user can specify whether
they want to list any additional publishing applications.

Therefore, this label is updated to be less generic.

### Add access to Whitehall question
As the "additional comments" question is quite generic, 2nd line content is
having to reply to users to specify whether the new user need access to
Whitehall publisher.

Add a new question that specifically asks whether the user needs access to
Whitehall publisher.

### Add access to other apps question
As the "additional comments" question is quite generic, 2nd line content is
having to reply to users to specify whether the new user need access to
applications other than Whitehall publisher.

Add a new question that specifically asks whether the user needs access to apps
other than Whitehall publisher.

### Additional comments can be blank
Additional comments is used to list the applications that the user needs access
to.

Now that we have a new question asking whether the user needs access to other
applications, this can be blank.

---

### DIFF
--- a/app/views/create_new_user_requests/_request_details.html.erb
+++ b/app/views/create_new_user_requests/_request_details.html.erb
@@ -74,14 +74,14 @@
     <div class="form-group">
       <span class="form-label">
         <%= f.label :additional_comments do %>
-          List any other publishing applications and permissions the user needs. If you’re not sure what these are, explain what tasks they need to be able to do.<abbr title="required">*</abbr>
+          List any other publishing applications and permissions the user needs. If you’re not sure what these are, explain what tasks they need to be able to do.
         <% end %>
       </span>
       <p class="help-block hint-block--md-6" id="additional-comments-hint">
         All users have basic access to Content Data, Feedback Explorer, GovSearch and Support forms.
       </p>
       <span class="form-wrapper">
-        <%= f.text_area :additional_comments, required: true, aria: { required: true }, class: "input-md-6 form-control", rows: 6, cols: 50 %>
+        <%= f.text_area :additional_comments, class: "input-md-6 form-control", rows: 6, cols: 50 %>
       </span>
     </div>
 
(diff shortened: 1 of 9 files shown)


app/views/create_new_user_requests/_request_details.html.erb
```diff
@@ -35,17 +35,53 @@
...
         <%= f.label :additional_comments do %>
-          What additional publishing permissions does the user need?<abbr title="required">*</abbr>
+          List any other publishing applications and permissions the user needs. If you’re not sure what these are, explain what tasks they need to be able to do.
         <% end %>
       </span>
       <p class="help-block hint-block--md-6" id="additional-comments-hint">
         All users have basic access to Content Data, Feedback Explorer, GovSearch and Support forms.
       </p>
       <span class="form-wrapper">
-        <%= f.text_area :additional_comments, required: true, aria: { required: true }, class: "input-md-6 form-control", rows: 6, cols: 50 %>
```
jkempster34, line 48: I've asked J-P whether this should be conditional validation based on the answer to the previous quesiton

jkempster34: It should, made that change and squashed it into the 4th commit

spec/features/create_new_user_requests_spec.rb
```diff
@@ -63,7 +63,7 @@
     fill_in "User's name", with: "Bob Fields"
     fill_in "User's email", with: "bob@gov.uk"
     select "Cabinet Office (CO)", from: "User's organisation"
-    fill_in "What additional publishing permissions does the user need?", with: "XXXX"
+    fill_in "List any other publishing applications and permissions the user needs. If you’re not sure what these are, explain what tasks they need to be able to do.", with: "XXXX"
```
brucebolt, line 74: I wonder if this field could be selected by id, rather than label text, to remove the need for the same text to be present in two places?  The [`fill_in` method](https://rubydoc.info/github/jnicklas/capybara/Capybara/Node/Actions:fill_in) appears to support this.

brucebolt (approved):

brucebolt (approved):

PR #1300, opened 2024-03-14
Support modern authentication methods for S3. 

sengi:
Essentially the same as https://github.com/alphagov/support-api/pull/911.

Remove Fog. The standard AWS client library is simpler to use, better documented and easier to secure. This lets us get rid of the nasty long-lived IAM key for authenticating to S3.


sengi:
Simplified the test fixture a bit more.

robinjam (approved):
Nice! Much cleaner than before as well

PR #1307, opened 2024-03-19
Update technical fault options as per March 2024 review

ChrisBAshton:
On @nicholsj' suggestion:

- We should have a "Search" option for general tech support issues (as opposed to the existing https://support.publishing.service.gov.uk/report_an_issue_with_govuk_search_results_request/new form, which is for reporting search results themselves)
- We should have an option for "Feedback explorer" (it's part of the Support app, but users don't necessarily know that, and it may well move into Content Data in future)
- We can make it clearer that "Do not know" is also for times where the user _does_ know the area, but it doesn't fit into any of the existing options.

NB, I'll get User Support to make sure the `fault_with_feedback_explorer` tag routes to [GOV.UK Platform Support](https://govuk.zendesk.com/agent/filters/12863141605916).

### DIFF
--- a/app/models/support/requests/technical_fault_report.rb
+++ b/app/models/support/requests/technical_fault_report.rb
@@ -27,12 +27,14 @@ module Support
         "content_tagger" => "Content Tagger",
         "datagovuk" => "data.gov.uk",
         "email_alerts" => "Email alerts",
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
Request for user support: https://govuk.zendesk.com/agent/tickets/5662004

PR #1320, opened 2024-04-08
Improve URL regex match

MahmudH:
Escaped `.` regular expression is more specific and restrictive in its matching of the domain name "gov.uk" by ensuring that the dot character is exactly matched, whereas the old regular expression allows for any character to be present in place of the dot. 

This means it will only match "gov.uk" and won't match variations like "govXuk" or "gov1uk".

Trello: https://trello.com/c/tiWPB7lJ/3474-review-code-scanning-alerts-for-our-repos

### DIFF
--- a/app/presenters/scope_filters_presenter.rb
+++ b/app/presenters/scope_filters_presenter.rb
@@ -81,7 +81,7 @@ private
       normalized_path = URI.parse(path_or_url).path
 
       if normalized_path.present?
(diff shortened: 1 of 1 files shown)


sengi (approved):
This fixes the (theoretical) issue with the `.`, but it'd probably be better still to get rid of these regexes altogether and use the [standard library's URL parser implementation](https://docs.ruby-lang.org/en/3.3/URI.html) rather than rolling our own.

There are other problems with these regexes besides the `.` vs `\.` thing, like they don't handle protocol-relative URLs and a whole bunch of other syntax.

PR #1332, opened 2024-04-22
Improve URL regex match

MahmudH:
Escaped . regular expression is more specific and restrictive in its matching of the domain name "gov.uk" by ensuring that the dot character is exactly matched, whereas the old regular expression allows for any character to be present in place of the dot.

This means it will only match "gov.uk" and won't match variations like "govXuk" or "gov1uk".

Trello: https://trello.com/c/tiWPB7lJ/3474-review-code-scanning-alerts-for-our-repos

### DIFF
--- a/app/models/support/requests/anonymous/explore.rb
+++ b/app/models/support/requests/anonymous/explore.rb
@@ -33,7 +33,7 @@ module Support
         def paths_from_urls
           urls = []
           parsed_urls.each do |url|
-            path = URI(url).path.sub(/^(http(s)?(:)?(\/)+?(:)?)?((\/)?www.)?gov.uk/, "")
+            path = URI(url).path.sub(/\A(http(s)?(:)?(\/)+?(:)?)?((\/)?www\.)?gov\.uk/, "")
             urls << (path.start_with?("/") ? path : "/#{path}")
           end
           urls.uniq


AgaDufrat (approved):
Thanks. This improves the Regex. 

I'm not sure why we need it in the first place. [URI `path` method](https://ruby-doc.org/stdlib-2.5.6/libdoc/uri/rdoc/URI/Generic.html) should already extract what we need here.

app/models/support/requests/anonymous/explore.rb
```diff
@@ -33,7 +33,7 @@ def redirect_path
         def paths_from_urls
           urls = []
           parsed_urls.each do |url|
-            path = URI(url).path.sub(/^(http(s)?(:)?(\/)+?(:)?)?((\/)?www.)?gov.uk/, "")
+            path = URI(url).path.sub(/\A(http(s)?(:)?(\/)+?(:)?)?((\/)?www\.)?gov\.uk/, "")
             urls << (path.start_with?("/") ? path : "/#{path}")
           end
           urls.uniq
```
AgaDufrat, line 39: I think the below _(not tested)_ should return exactly the same thing.
``` suggestion
          parsed_urls.map { |url| URI(url).path }.uniq
```

PR #1536, opened 2024-11-21
Upgrade app to Rails 7.2

MuriloDalRi:
Upgrades Rails from 7.1 to 7.2


config/environments/development.rb
```diff
@@ -36,8 +34,12 @@
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
 
+  # Disable caching for Action Mailer templates even if Action Controller
+  # caching is enabled.
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
```
AgaDufrat, line 41: I think we tend to use Plek for those. Surprised it didn't suggest updating those in other environments.

AgaDufrat (approved):

PR #1665, opened 2025-04-01
Update create new user request

yndajas:
[Trello](https://trello.com/c/IN20Asaz/1634-updates-to-the-create-new-user-support-form)

This updates the create new user request form per a content team request. The main changes are two new questions, minor tweaks to existing copy, and fieldsets for grouping questions

I haven't updated the name of the files or classes here in order to keep things focused on the user-facing changes with minimal diff, but I have opened #1666 to do that as an optional extra

### Before

2025-04-01 17 57 38 support dev gov uk 323fb116c394

### After

2025-04-02 08 56 30 support dev gov uk d6e084036fcc

### DIFF
--- a/app/views/create_new_user_requests/_request_details.html.erb
+++ b/app/views/create_new_user_requests/_request_details.html.erb
@@ -116,6 +116,27 @@
       </span>
     </div>
 
+    <div class="form-group">
+      <span class="form-label">
+        <%= f.label :writing_for_govuk_training, class: "control-label" do %>
+          Does the user require Writing for GOV.UK training?
+        <% end %>
+      </span>
+      <p class="help-block hint-block--md-6" id="writing-training-hint">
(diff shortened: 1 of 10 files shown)


app/views/create_new_user_requests/_request_details.html.erb
```diff
@@ -38,6 +38,24 @@
       </span>
     </div>
 
+    <div class="form-group">
+      <span class="form-label">
+        <%= f.label :new_or_existing_user, class: "control-label" do %>
```
brucebolt: Probably out of scope for this PR, but I wonder if the questions could also be added into the configuration file, so everything is configured there?

brucebolt (approved):
I'm not particularly familiar with this app, but all the changes looks good to me.

app/views/create_new_user_requests/_request_details.html.erb
yndajas: Feels sensible - there or the en.yml file. But yeah, maybe one for the backlog to do that here and in other forms

PR #1781, opened 2025-08-06
[PP-7296] Design systemise change user form

yndajas:
- Add support for using the Design System for request forms (all but the last commit)
- Use the Design System for the "Change an existing user's account" form (final commit)

The former could be its own PR, but it probably makes more sense with the added context of the final commit

## Screen captures

image

https://github.com/user-attachments/assets/fe4819e2-5ddb-4e00-872c-a7ab80e5dc52


callumknights (approved):
Happy with all of this 👍

PR #2043, opened 2026-06-19
Upgrade js-yaml for dependabot #164

nacnudus:
This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1925,7 +1925,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"js-yaml@npm:^4.1.0":
+"js-yaml@npm:^4.1.0, js-yaml@npm:^4.2.0":
   version: 4.2.0
   resolution: "js-yaml@npm:4.2.0"
   dependencies:
@@ -3339,6 +3339,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "support@workspace:."
   dependencies:
+    js-yaml: "npm:^4.2.0"
     standardx: "npm:^7.0.0"
     stylelint: "npm:^17.13.0"
     stylelint-config-gds: "npm:^3.0.0"
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "extends": "stylelint-config-gds/scss"
   },
   "devDependencies": {
+    "js-yaml": "^4.2.0",
     "standardx": "^7.0.0",
     "stylelint": "^17.13.0",
     "stylelint-config-gds": "^3.0.0"


yarn.lock
```diff
@@ -1925,7 +1925,7 @@ __metadata:
   languageName: node
   linkType: hard
```
brucebolt, line 1926: It looks like `3.14.2` is still being included elsewhere, because of another dependency.

I'm pretty sure this alert can't be resolved as we have a dependency tree that results in requiring eslintrc `0.3.0` which [depends on js-yaml `3.13.1`](https://github.com/eslint/eslintrc/blob/cccaad4b3fb14e272aa4aa42a5b60750e03f80ea/package.json#L54). We'll have to wait for this to be resolved before we can fully upgrade.

yndajas: We might be waiting a while for that. It looks like the dependency tree is:
- [standardx 7](https://github.com/standard/standardx/blob/v7.0.0/package.json)
- [standard 16](https://github.com/standard/standard/blob/16.x/package.json)
- [eslint 7.18](https://github.com/eslint/eslint/blob/v7.18.0/package.json)
- js-yaml 3

standardx hasn't seen an release since 2020, so the likelihood of a timely bump seems low 🥲

We could try overriding the version of standard via a [resolution](https://yarnpkg.com/configuration/manifest#resolutions), but [it sounds like standardx 7 doesn't play nicely with standard 17](https://github.com/standard/standardx/issues/44)

I think a better approach would be to remove the standardx dependency and migrate to standard and/or eslint. Some notes:
- [the commit that introduced standardx](https://github.com/alphagov/support/commit/93d0439b9eaee4d2bc08cf02c93b7ec4fe0d9aea) explained that it was to override a rule to support IE11, [which is no longer in the list of browsers to test against](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-february-2026), and [the author of the introducing commit has since said this isn't needed](https://github.com/alphagov/govuk-chat/commit/b134b8efccce453261275ab3255a9dae1411e080)
- there's prior art in Whitehall and govuk-chat:
  - https://github.com/alphagov/whitehall/pull/8178
  - https://github.com/alphagov/govuk-chat/commit/b134b8efccce453261275ab3255a9dae1411e080
  - https://github.com/alphagov/govuk-chat/commit/dc6590a7115b9871772b583bb3c683c246afa321
- [the shared CI workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/standardx.yml) only supports standardx but the prior art covers running standard/eslint in CI

yndajas: (the eslintrc 0.3.0 dependency comes from eslint 7.18, so it's a branch of the same tree)

brucebolt:
Marking draft, as this is covered by https://gov-uk.atlassian.net/browse/PP-7810.

PR #2055, opened 2026-07-07
Sch 1997 improve the search support form

emmalowe:
Changes to the support form for search to make it more useful and usable. 

See JIRA ticket for requirements: https://gov-uk.atlassian.net/browse/SCH-1997 

This is how the new version of the form looks:

Screenshot 2026-07-14 at 11 27 56

### DIFF
--- a/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
+++ b/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
@@ -5,7 +5,6 @@ module Support
 
       validates :search_query, presence: true
       validates :results_problem, presence: true
-      validates :change_justification, presence: true
 
       def self.label
         "Report an issue with GOV.UK search results"
--- a/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
+++ b/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
@@ -1,7 +1,7 @@
 module Support
   module Requests
     class ReportAnIssueWithGovukSearchResultsRequest < Request
-      attr_accessor :search_query, :results_problem, :change_requested, :change_justification, :evidence_availability
+      attr_accessor :search_query, :results_problem, :change_requested, :change_justification, :evidence_availability, :evidence_description
 
       validates :search_query, presence: true
       validates :results_problem, presence: true
--- a/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
+++ b/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
@@ -5,7 +5,6 @@ module Support
 
       validates :search_query, presence: true
       validates :results_problem, presence: true
-      validates :change_requested, presence: true
       validates :change_justification, presence: true
 
       def self.label
--- a/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
+++ b/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
@@ -13,7 +13,7 @@ module Support
       end
 
       def self.description
-        "Report a problem with specific search results from the new GOV.UK search engine."
+        "Report a problem or request a change to search results on GOV.UK."
       end
     end
   end
--- a/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
+++ b/app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
@@ -1,10 +1,23 @@
 module Support
   module Requests
     class ReportAnIssueWithGovukSearchResultsRequest < Request
-      attr_accessor :search_query, :results_problem, :change_requested, :change_justification
+      attr_accessor :search_query, :results_problem, :change_requested, :change_justification, :evidence_availability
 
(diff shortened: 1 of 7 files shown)


app/models/support/requests/report_an_issue_with_govuk_search_results_request.rb
```diff
@@ -1,19 +1,30 @@
...
   module Requests
     class ReportAnIssueWithGovukSearchResultsRequest < Request
-      attr_accessor :search_query, :results_problem, :change_requested, :change_justification
+      attr_accessor :search_query, :results_problem, :change_requested, :change_justification, :evidence_availability, :evidence_description
 
       validates :search_query, presence: true
       validates :results_problem, presence: true
-      validates :change_requested, presence: true
-      validates :change_justification, presence: true
+      validates :evidence_availability, inclusion: { in: %w[yes no not_sure] }
```
emmalowe, line 8: I've based these changes off what already exists in the taxonomy change request form:

https://github.com/alphagov/support/blob/main/app/models/support/requests/taxonomy_change_topic_request.rb

app/views/report_an_issue_with_govuk_search_results_requests/_request_details.html.erb
```diff
@@ -32,22 +35,54 @@
...
-        Explain why this change is necessary, including the impact on users and the number of users affected.
+        If applicable, explain why this change is necessary. Why are the current search results bad for users?
+      <% end %>
+    </span>
+    <span class="form-wrapper">
+      <%= f.text_area :change_justification, class: "input-md-6 form-control", rows: 6, cols: 50 %>
+    </span>
+  </div>
+
+  <div class="form-group">
```
emmalowe, line 60: See taxonomy change example: https://github.com/alphagov/support/blob/main/app/views/taxonomy_change_topic_requests/_request_details.html.erb#L16

spec/models/support/requests/report_an_issue_with_govuk_search_results_request_spec.rb
```diff
@@ -3,10 +3,25 @@
 module Support
   module Requests
     describe ReportAnIssueWithGovukSearchResultsRequest do
+      def request(options = {})
```
emmalowe, line 6: See taxonomy change example: https://github.com/alphagov/support/blob/main/spec/models/support/requests/taxonomy_change_topic_request_spec.rb

app/views/report_an_issue_with_govuk_search_results_requests/_request_details.html.erb
```diff
@@ -1,27 +1,30 @@
...
   <div class="form-group">
     <span class="form-label">
       <%= f.label :search_query do %>
-        What search query did you use, for example “Universal Credit”?
-      <% end %>
+        What search queries are not working well?<abbr title="required">*</abbr>
+    <% end %>
     </span>
     <span class="form-wrapper">
+      <span class="help-block">
```
hannako: It looks like you've got a span nested inside a span. I think we could consolidate to just a single span : 

`<span class = "form-wrapper help-block hint-block--md-6>`

The `hint-block--md-6` class is keeping the text nicely sitting on top of the form field. I spotted it in a few places eg https://github.com/alphagov/support/blob/16b1b9232615aa15c1b5e3838251ebdec24c1af6/app/views/content_change_requests/_request_details.html.erb#L13

app/views/report_an_issue_with_govuk_search_results_requests/_request_details.html.erb
```diff
@@ -32,22 +35,54 @@
   <div class="form-group">
     <span class="form-label">
       <%= f.label :change_requested do %>
-        What change do you want to make to the search results? Include URLs for any pages you want the search to show.
+        If applicable, which pages are missing, or showing too high or low in results? If the pages are showing do you think they should be higher, lower, or removed?
```
hannako, line 38: I think we might also want to apply the class to here too eg

`<%= f.label :change_requested, class: "hint-block--md-6" do %>`

hannako (commented):
This looks great, my only comments are around the styling. It would be much easier if this app had been migrated to the design system!

app/views/report_an_issue_with_govuk_search_results_requests/_request_details.html.erb
emmalowe: Ah yes, good point. I was following a different example, but I agree that this is better.

Nice find on the wrap styling!

hannako (commented):
The changes look good to me based on the existing patterns in this repo - I'll leave it to the owning team to approve the PR.

brucebolt (approved):
Everything looks good to me 👍.

Thanks for breaking the PR down into small commits, which made it very easy to review.

brucebolt:
> It would be much easier if this app had been migrated to the design system!

Agreed. We have [a ticket](https://gov-uk.atlassian.net/browse/PP-7801) to migrate the app to the design system. But in reality, we won't be working on this anytime soon as this is considered a "Level 2" app by our team.

PR #2097, opened 2026-08-27
[PP-7957] Add Content API to technical fault report form

brucebolt:
This will allow users to report issues with the Content API.

These tickets will be assigned to `GOV.UK 2nd Line--Alerts and Issues` until we ask a Zendesk admin to set up a trigger for tickets tagged with `fault_with_content_api`.

As an aside, I've also removed Maslow from the list as this app is no longer available to publishers.

### DIFF
--- a/app/models/support/requests/technical_fault_report.rb
+++ b/app/models/support/requests/technical_fault_report.rb
@@ -21,6 +21,7 @@ module Support
       # ("Conditions" -> "Tags" -> "Contains at least one of the following")
       OPTIONS = {
         "collections_publisher" => "Collections Publisher",
+        "content_api" => "Content API",
         "content_data" => "Content Data",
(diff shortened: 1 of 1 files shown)


callumknights (commented):
Feels really nitpicky. But does the Maslow removal belong in this PR? Especially because the PR is linked to a Jira task

brucebolt:
> Feels really nitpicky. But does the Maslow removal belong in this PR? Especially because the PR is linked to a Jira task

You're absolutely right, I was just being lazy. I'll raise another PR to remove Maslow.

brucebolt:
I've raised https://github.com/alphagov/support/pull/2098 for Maslow.

callumknights (approved):
Thanks! 👍

PR #2102, opened 2026-09-01
[PP-7948] Bump govuk_publishing_components 67.0.1 -> 68.3.0

yndajas:
v68 bumps govuk-frontend (a Node package) to v6. The only thing that appears to affect us here is the removal of layout-header.js, but we weren't relying on it (and weren't even loading it until the very recent 3c5f103b1cab89b03471a81acbe160bd06e2904f)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -210,7 +210,7 @@ GEM
     govuk_personalisation (1.2.0)
       plek (>= 1.9.0)
       rails (>= 6, < 9)
-    govuk_publishing_components (67.0.1)
+    govuk_publishing_components (68.3.0)
       chartkick
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -178,13 +178,13 @@ GEM
       warden-oauth2 (~> 0.0.1)
     globalid (1.4.0)
       activesupport (>= 6.1)
-    google-protobuf (4.35.1-aarch64-linux-gnu)
+    google-protobuf (4.36.1-aarch64-linux-gnu)
```
callumknights: Lots of other dependencies here, were they added by accident with the upgrade?

yndajas: Ha, I added an inverse comment: https://github.com/alphagov/transition/pull/2291#discussion_r3906382760

What command did you use?

callumknights: `bundle update govuk_publishing_components --conservative`

yndajas: Updated per the other thread

callumknights (approved):

PR #2115, opened 2026-09-17
[PP-7990] Prevent CookieOverflow errors

geek4good:
We've seen some CookieOverflow error on Sentry (e.g. https://govuk.sentry.io/issues/7730616916/), which have been caused by overly long and clearly bogus Referer URLs.

This change introduces a middleware that simply checks if the referer and query string of a request are no longer than 2048 bytes, otherwise it discards them.


spec/requests/omni_auth_request_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    get "/auth/gds", headers: { "Referer" => "https://support.publishing.service.gov.uk/anonymous_feedback" }
+
+    expect(response).to have_http_status(:redirect)
+  end
+
+  it "starts sign-in normally for an ordinary query string" do
+    get "/auth/gds?this=is-fine"
+
+    expect(response).to have_http_status(:redirect)
+  end
```
yndajas, line 35: **suggestion:** I'm assuming the response location will reflect the referrer and the query string? Is it worth asserting on those for documentation?

config/application.rb
```diff
@@ -12,6 +12,7 @@
 # require "action_text/engine"
 require "action_view/railtie"
 # require "action_cable/engine"
+require_relative "../lib/middleware/omni_auth_request_sanitizer"
```
yndajas: **nitpick:** I think we'd usually put all the `require`s before or after all the `require_relative`s, but it's a bit of a mixed bag across the org: https://github.com/search?q=org%3Aalphagov+require_relative&type=code

lib/middleware/omni_auth_request_sanitizer.rb
```diff
@@ -0,0 +1,18 @@
+module Middleware
+  class OmniAuthRequestSanitizer
```
yndajas, line 2: **thought:** I quite like this approach. It's worth mentioning that we've addressed this in a different way over on Signon and gds-sso (gem):
- https://github.com/alphagov/signon/commit/79705b04a8b93c99fae7bca0d176b3eedd03ed53
- https://github.com/alphagov/gds-sso/commit/c07105927a014f9468fb5f03730e219b2358cc69

I don't have any immediate thoughts on whether one way is better than the other - it looks like your approach might be more precise, whereas the other approach was just looking at the path length?

yndajas (commented):
Nothing really blocking, but a few quick thoughts

lib/middleware/omni_auth_request_sanitizer.rb
geek4good: I was hoping you'd be the one to review this, exactly because you've fixed this elsewhere in a different way.

I'm not sure about the precision of the approach, but it seemed like a good idea to try and identify bogus requests / headers as early as possible and neutralise them "at the door" instead of having to deal with them later. Luckily, the error that prompted this particular fix, it was an overly long Referer string, so we could add that to the list of possible causes of things blowing up.

Would it be over the top to extract the middleware and use it in all the apps where this could happen?

yndajas: That sounds pretty useful to me. I'm not sure if there's an obvious gem for it to live in? govuk_app_config might be the closest given it's quite generic, but I don't _think_ it typically houses middleware. There's always the option of creating a new gem, but this is quite a small amount of code 🤔

Would this middleware approach work for the gds-sso use case, or could it replace it?

geek4good: I think this could replace the gds-sso use case, which was due to an overly long redirect path, if I'm not misreading it. This middleware should be able to handle that. Might be worth iterating over the query string and only removing key-value pairs that meet certain criteria, instead of completely removing the query string in case it's too long.

I don't think govuk_app_config would be a bad place to put something like this. Might be worth a discussion in our next dev chat.

spec/requests/omni_auth_request_spec.rb
geek4good: Took me a while to figure out, but this should now demonstrate it working for "good" and "bad" data.

geek4good: I started decrypting cookies at some point, but it good really weird really fast.

lib/middleware/omni_auth_request_sanitizer.rb
yndajas: Yeah that sounds good to me. I think we could merge this as is then discuss an abstraction separately

yndajas (approved):
Looks great

PR #2116, opened 2026-09-18
[PP-7991] Gracefully handle unknown organisation slugs

geek4good:
When presented with an unknown organisation slug, the `ScopeFiltersPresenter` used to raise a GdsApi::HTTPNotFound error mid-render, which led to an `ActionView::Template::Error` showing up in Sentry (e.g. https://govuk.sentry.io/issues/7730848769/). This change catches the error and falls back to an organisation name derived from the organisation slug.


app/presenters/scope_filters_presenter.rb
```diff
@@ -49,7 +49,12 @@ def paths_title
   end
 
   def organisation
-    @organisation ||= Services.support_api.organisation(organisation_slug) if organisation_slug.present?
+    return if organisation_slug.blank?
+
+    @organisation ||= Services.support_api.organisation(organisation_slug)
+  rescue GdsApi::HTTPNotFound
+    # Fall back to a title derived from the slug instead of raising mid-render
+    @organisation = { "title" => organisation_slug.tr("-", " ").humanize }
```
yndajas, line 57: **thought:** as far as I can see:
- the title will be used for the h1, page title, and breadcrumbs
- the organisation select element will be blank, since the slug won't match any of the values
- presumably there will be no results

The slug is used all over the place, but I can't see that having an incorrect one would cause an issue anywhere other than here

I wonder if in this situation, we should be showing a flash message to say that no organisation has been found with the given slug and to check the URL or try selecting an organisation again? We'd probably only want to set this on the anonymous feedback index though, not on the create action nor the export requests controller which also use the scope filters. So I'm not sure how easy it would be to wire that up 🤔 maybe the call to Support API belongs outside the presenter

We'd still need to put something in the h1, page title, and breadcrumbs, so maybe this fake title would still be beneficial for those use cases

What do you reckon?

geek4good: Great feedback. I wasn't really sure what to make of this whole thing, and you've helped me take a step back and rewrite the controller and presenter in a way that makes a lot more sense in my opinion.

spec/controllers/anonymous_feedback/export_requests_controller_spec.rb
```diff
@@ -110,6 +110,41 @@
                                         ))
       end
     end
+
+    context "when the organisation slug isn't know to support-api" do
```
yndajas: **typo:** "know" -> "known"

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -287,18 +240,8 @@
...
-        "department-of-hats",
-        slug: "department-of-hats",
-        web_url: "https://www.gov.uk/government/organisations/department-of-hats",
-        title: "Department of Hats",
-        acronym: "DoH",
-        govuk_status: "live",
-      )
+    let(:organisation) { { "title" => "Department of Hats" } }
 
-      stub_support_api_anonymous_feedback_doc_type_summary(document_type: "smart_answer")
```
yndajas: **question:** what allows `stub_support_api_anonymous_feedback_doc_type_summary` to be removed?

spec/presenters/scope_filters_presenter_spec.rb
```diff
@@ -187,64 +187,17 @@
...
+        organisation: { "title" => "Department of Hats" },
       )
-      presenter = described_class.new(organisation_slug: "department-of-hats")
       expect(presenter.organisation_title).to eq "Department of Hats"
     end
 
-    it "is nil if no organisation_slug was provided" do
-      presenter = described_class.new(organisation_slug: nil)
+    it "is nil if no organisation was found" do
+      presenter = described_class.new(organisation_slug: nil, organisation: nil)
```
yndajas: **suggestion:** I wonder if the descriptions are a little too aware of other actors? As far as this class is concerned, it only cares whether an organisation (with a title) was passed in

E.g.:
- "retrieves the title from the organisation when passed in"
- "is nil if no organisation was provided"

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -91,9 +93,22 @@ def scope_filters
...
+  def find_organisation
+    slug = index_params[:organisation]
+    return if slug.blank?
+
+    Services.support_api.organisation(slug)
+  rescue GdsApi::HTTPNotFound
+    flash.now[:warning] = "We couldn't find an organisation with slug '#{slug}'. " \
+      "Check the URL, or choose an organisation from the list below."
+    # Fall back to a title derived from the slug
+    { "title" => "Organisation not found" }
```
yndajas: **thought:** I like this, but I still wonder if these title/h1/breadcrumbs that it would produce quite read right:
- `Feedback for “Organisation not found on / - Document type: about” | GOV.UK Support`
- `Organisation not found on / - Document type: about`

It's a tricky one though and I don't have an alternative suggestion 😅 so I think it's probably okay for an unhappy path

yndajas (commented):
This looks pretty good - nothing really blocking here, just a few small thoughts/questions/nits

spec/presenters/scope_filters_presenter_spec.rb
geek4good: This was a happy accident. Removed it unintentionally when moving some stuff around and then noticed it was apparently redundant.

yndajas: Oh nice. I'd probably do that as it's own commit to make it clear it's unrelated

app/controllers/anonymous_feedback_controller.rb
geek4good: I was wondering if the following might be better than setting the title to some random thing:
```ruby
rescue GdsApi::HTTPNotFound
  flash[:warning] = "We couldn't find an organisation with slug '#{slug}'. " \
    "Check the URL, or choose an organisation from the list below."
  redirect_back_or_to anonymous_feeedback_explore_path
end
```
What do you think?

yndajas: I wondered that too and then totally forgot! I think that makes sense. I guess the only risk is losing any other params the user has set (e.g. in a copy paste error). "Check the URL" probably wouldn't apply on the explore path, but it might be a better flow 🤔

app/controllers/anonymous_feedback_controller.rb
```diff
@@ -91,9 +93,22 @@ def scope_filters
       path_set_id: saved_paths.try(:id),
       organisation_slug: index_params[:organisation],
       document_type: index_params[:document_type],
+      organisation: @organisation,
     )
   end
 
+  def find_organisation
```
yndajas: **thought:** I forgot another thing - given this is only used for the index, and has logic specifically intended for the index (the flash, and possibly the [redirect](https://github.com/alphagov/support/pull/2116/changes#r4085122402)), I wonder if it's worth naming it more explicitly (e.g. `find_organisation_for_index`) to warn/prevent future users?

Alternatively the `rescue` could move into the index action, but then the method still uses `index_params`

